PR #7, opened 2023-05-28
Setup DB

### LINKED ISSUE #2: Setup DB
After #1, we should setup the app's DB, perhaps as another Docker Compose service.

Needed tasks include:
- [ ] Investigate possible DB solutions;
- [ ] Build DB service;
- [ ] Update App config to reflect the changes;

After these changes are done, migrations should be ran.

ttoino:
Closes #2.

Adds a new `postgres` service and adds the relevant configuration items to the `.env` file.

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,12 @@
+APP_NAME="SINF Website 2023"
+APP_ENV=local
+APP_KEY=base64:AVegBkyjHs2QQsx53M9Kt9GbnrDLc6yG4cRg+WMJrbA=
+APP_DEBUG=true
+APP_URL=http://localhost
+
+DB_HOST=localhost
+DB_PORT=5432
+DB_USERNAME=postgres
+DB_PASSWORD=postgres
+DB_DATABASE=sinf-website-2023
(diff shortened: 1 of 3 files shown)


.env
```diff
@@ -0,0 +1,12 @@
...
+APP_KEY=base64:AVegBkyjHs2QQsx53M9Kt9GbnrDLc6yG4cRg+WMJrbA=
+APP_DEBUG=true
+APP_URL=http://localhost
+
+DB_HOST=localhost
+DB_PORT=5432
+DB_USERNAME=postgres
+DB_PASSWORD=postgres
+DB_DATABASE=sinf-website-2023
+DB_CONNECTION=pgsql
```
Naapperas, line 12: I'm not entirely sure about these being made public: any ideas/thoughts?

Naapperas (changes requested):

.env
ttoino: These are meant mostly for testing locally, production values would be in a separate file

Naapperas (approved):

PR #13, opened 2023-07-04
Create database schema

### LINKED ISSUE #12: Create the database SQL file
Related to #2, we now have a UML representation of the database, as such we should start creating the SQL file for the database.
Below is the UML.

SINF website-ConceptualModel, v2 drawio

Top priority should be the Auth, Activity and SINF sections.

ttoino:
Closes #12.

I created a migration to create all the needed tables.

Still to do:

- [x] Finish models (right now they are all empty)
- [x] Add quest unlock conditions
- [x] Add `onDelete`s and `onUpdate`s to foreign keys
- [ ] Add `CHECK` constraints (if needed)

### DIFF
--- a/app/Models/Enrollment.php
+++ b/app/Models/Enrollment.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+use Illuminate\Database\Eloquent\Model;
+
+class Enrollment extends Model
+{
+    use HasFactory;
+}
--- a/app/Models/Enrollment.php
+++ b/app/Models/Enrollment.php
@@ -4,8 +4,37 @@ namespace App\Models;
 
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 
-class Enrollment extends Model
-{
+class Enrollment extends Model {
     use HasFactory;
+
+    /**
+     * The attributes that are mass assignable.
+     *
+     * @var array<int, string>
+     */
+    protected $fillable = [
+    ];
+
+    public function edition(): BelongsTo {
+        return $this->belongsTo(Edition::class);
+    }
+
+    public function events(): BelongsToMany {
+        return $this->belongsToMany(Event::class);
+    }
+
+    public function products(): BelongsToMany {
+        return $this->belongsToMany(Product::class);
+    }
+
+    public function quests(): BelongsToMany {
+        return $this->belongsToMany(Quest::class);
+    }
+
+    public function student(): BelongsTo {
+        return $this->belongsTo(Student::class);
+    }
 }
--- a/database/migrations/2023_07_04_010134_create_schema.php
+++ b/database/migrations/2023_07_04_010134_create_schema.php
@@ -92,7 +92,8 @@ return new class extends Migration {
             // we need to be careful to update this when we add new categories
             $table->enum('category', ['COMPANY', 'TALK', 'WORKSHOP', 'MILESTONE', 'TEAMBUILDING']);
 
-            // TODO: Unlock condition
+            $table->morphs('redeemable');
+
             $table->foreignIdFor(\App\Models\Edition::class)->constrained()->cascadeOnDelete();
             $table->timestamps();
         });
@@ -157,6 +158,7 @@ return new class extends Migration {
      * Reverse the migrations.
      */
     public function down(): void {
+        Schema::dropIfExists('social_media');
         Schema::table('users', function (Blueprint $table) {
             $table->dropMorphs('usertype');
         });
--- a/database/migrations/2023_07_04_010134_create_schema.php
+++ b/database/migrations/2023_07_04_010134_create_schema.php
@@ -15,7 +15,8 @@ return new class extends Migration {
             $table->morphs('usertype');
         });
         Schema::create('students', function (Blueprint $table) {
-            $table->foreignIdFor(\App\Models\User::class)->primary()->constrained()->cascadeOnDelete();
+            $table->id();
+            $table->foreignIdFor(\App\Models\User::class)->unique()->constrained()->cascadeOnDelete();
             $table->string('linkedin')->nullable();
             $table->string('twitter')->nullable();
             $table->string('facebook')->nullable();
@@ -24,11 +25,13 @@ return new class extends Migration {
             $table->timestamps();
         });
         Schema::create('companies', function (Blueprint $table) {
-            $table->foreignIdFor(\App\Models\User::class)->primary()->constrained()->cascadeOnDelete();
+            $table->id();
+            $table->foreignIdFor(\App\Models\User::class)->unique()->constrained()->cascadeOnDelete();
             $table->timestamps();
         });
         Schema::create('admins', function (Blueprint $table) {
-            $table->foreignIdFor(\App\Models\User::class)->primary()->constrained()->cascadeOnDelete();
+            $table->id();
+            $table->foreignIdFor(\App\Models\User::class)->unique()->constrained()->cascadeOnDelete();
             $table->timestamps();
         });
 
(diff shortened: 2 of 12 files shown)


database/migrations/2023_07_04_010134_create_schema.php
```diff
@@ -0,0 +1,127 @@
...
+        Schema::create('companies', function (Blueprint $table) {
+            $table->id();
+            $table->foreignIdFor(\App\Models\User::class);
+            $table->timestamps();
+        });
+        Schema::create('admins', function (Blueprint $table) {
+            $table->id();
+            $table->foreignIdFor(\App\Models\User::class);
+            $table->timestamps();
+        });
```
ttoino: I'm unsure if this is the best way to model these relations, here I'm just making a one to one relationship to the User model (already created before) and each of the user types, but this would allow each user to belong to multiple types and wouldn't let us easily find the user's type in code. An alternative I found is to use Eloquent's polymorphic relationships, but that would require adding two columns to the User model, one for the user type and another for the user type id. If you find other alternatives, feel free to share.

database/migrations/2023_07_04_010134_create_schema.php
```diff
@@ -0,0 +1,127 @@
...
+            $table->foreignIdFor(\App\Models\Event::class);
+            $table->timestamps();
+        });
+
+        // Quest tables
+        Schema::create('quests', function (Blueprint $table) {
+            $table->id();
+            $table->string('name');
+            $table->integer('points')->unsigned();
+            $table->enum('category', ['COMPANY', 'TALK', 'WORKSHOP', 'MILESTONE', 'TEAMBUILDING']);
```
Naapperas, line 94: Could you move this list elsewhere? Perhaps the Quest model could have this as a static element, this way we can future-proof this code.

database/migrations/2023_07_04_010134_create_schema.php
```diff
@@ -0,0 +1,127 @@
...
+        });
+
+        // Enrollment
+        Schema::create('enrollments', function (Blueprint $table) {
+            $table->id();
+            $table->foreignIdFor(\App\Models\Student::class);
+            $table->foreignIdFor(\App\Models\Edition::class);
+            $table->timestamps();
+            $table->unique(['student_id', 'edition_id']);
+        });
```
Naapperas: We are missing the `points` attribute in here.

Naapperas: Even though it is a derived attribute, it should still be an attribute, in my opinion.

Naapperas (changes requested):

database/migrations/2023_07_04_010134_create_schema.php
ttoino: I'll look into it, but I doubt it's the best idea since this is a migration and introducing dependencies like that doesn't sound like the best idea

database/migrations/2023_07_04_010134_create_schema.php
ttoino: Do you think that's the best approach? We can just use the Eloquent ORM to calculate the points

database/migrations/2023_07_04_010134_create_schema.php
Naapperas: I get your point. I just feel like this can bring a lot of trouble in the future due to people forgetting to update the list in both places but honestly I don't have a better idea.

database/migrations/2023_07_04_010134_create_schema.php
Naapperas: That is true, but I thought that perhaps we should keep track of the points a person acquired over the years. Perhaps we don't, in which case we could just use Eloquent to do that. Its a matter of deciding as a team.

database/migrations/2023_07_04_010134_create_schema.php
```diff
@@ -0,0 +1,179 @@
...
+        // Quest tables
+        Schema::create('quests', function (Blueprint $table) {
+            $table->id();
+            $table->string('name');
+            $table->integer('points')->unsigned();
+
+            // we need to be careful to update this when we add new categories
+            $table->enum('category', ['COMPANY', 'TALK', 'WORKSHOP', 'MILESTONE', 'TEAMBUILDING']);
+
+            $table->morphs('redeemable');
```
CiscoPr: I'm probably nitpicking, but if you're referring to the reward attribute on the Quest table of the UML, shouldn't you also name it as it is in the diagram? Again, it's a really insignificant detail, but still, I think we should remain 100% faithful to the names we gave in the UML
```suggestion
            $table->morphs('reward');
```

ttoino: That's not the reward, that's the event/company that needs to be redeemed to unlock the quest, redeemable may not be the best name tho, maybe requirement?

CiscoPr: yes, it's a good replacement in my opinion. At least it's a more impartial and straightforward name

Naapperas:
Take a look at #14 and #15. Those items have high priority (companies are joining on the premise that they will have access to the relevant info) and I believe that they can be implemented by this PR:
- #14 is a matter of adding an attribute for the tiers to the company schema, if I'm not mistaken
- #15 can be implemented through a new table that directly registers company visits or can be implemented through some queries to the *quest* table, we need to see what works best

@ttoino if you can take a look at these when you can, so we can close this PR ASAP and start work on the remaining features.

pedrosilva17:
About #15, I believe that introducing a bit of redundancy through a new "visit history" table (as Nuno said) is worth it in this case. Along with it, we could have an on-create trigger in the `enrollment_quest` table, which inserts a row in this new table (perhaps student_id, company_id and timestamps should suffice). 
Instead of running a fairly complex query every time a company manager wants to see who visited them, we pass that load onto the trigger, which runs once every time a quest of that category is completed by a student.

Naapperas:
Perhaps we should merge this into main already, so other work can be started. Then new PRs targetting the missing features could be made. Thoughts @ttoino ?

Naapperas (commented):
Do the relations on `Edition` and `Enrollment` work? I know there are some "through" variants but I do not see them used anywhere, and from looking at the migration code I do not see some of the relations defined, maybe I am missing something...

ttoino:
> Perhaps we should merge this into main already, so other work can be started. Then new PRs targetting the missing features could be made. Thoughts @ttoino ?

I think that's a good idea. The requirements for #14 should be done, but I think the requirements for #15 should be added in that issue's PR. Just keep in mind any further additions/modifications to the DB schema should be done in separate migrations, and this one should not be edited after merging.

ttoino:
> Do the relations on `Edition` and `Enrollment` work? I know there are some "through" variants but I do not see them used anywhere, and from looking at the migration code I do not see some of the relations defined, maybe I am missing something...

I can add some through relations if they are deemed necessary, but I don't think any are missing for the enrollment, we have enrollment \*--1 edition, enrollment \*--1 user, enrollment \*--\* quest (unlocked quests), enrollment \*--\* event (events enrolled in), enrollment \*--\* products (bought products).

Naapperas:
> > Do the relations on `Edition` and `Enrollment` work? I know there are some "through" variants but I do not see them used anywhere, and from looking at the migration code I do not see some of the relations defined, maybe I am missing something...
> 
> I can add some through relations if they are deemed necessary, but I don't think any are missing for the enrollment, we have enrollment *--1 edition, enrollment *--1 user, enrollment *--* quest (unlocked quests), enrollment *--* event (events enrolled in), enrollment *--* products (bought products).

Oh right, then it's all good.

Naapperas (approved):
We are just missing some `CHECK`s after this I believe, but great job.

app/Models/Enrollment.php
```diff
@@ -0,0 +1,40 @@
...
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+
+class Enrollment extends Model {
+    use HasFactory;
+
+    /**
+     * The attributes that are mass assignable.
+     *
+     * @var array<int, string>
+     */
```
CiscoPr, line 17: Will this fillable array have any attributes at all?

CiscoPr (commented):
Everything looks good me. I just have that question regarding the Enrollment model. Nevertheless, very good job @ttoino

app/Models/Enrollment.php
```diff
@@ -0,0 +1,40 @@
...
+    protected $fillable = [
+    ];
```
CiscoPr, line 19: will this fillable array have any attributes at all?

PR #16, opened 2023-08-04
File uploading

### LINKED ISSUE #9: Implement File Uploads
(Normal) Users should be able to upload their CVs.

This should be done through a file upload system.

pedrosilva17:
Closes #9
This PR implements a file upload system to use for CVs and potentially other stuff in the future.

Currently, it is very basic and one-dimensional, merely serving as a test to verify that the file uploads are handled correctly.

The idea is to create a generic middleware which will take care of redirecting to the appropriate controller, as well as making use of Jetstream Actions. The message I've sent on Slack is relevant to this, with a forum page and a video regarding the suggested approach to be adopted. Leave your suggestions below if you think of other ways to do this.


Naapperas:
Can you elaborate on what you mean by "redirecting to the appropriate controller"?

pedrosilva17:
I typed that after seeing this section of the Laravel documentation: https://laravel.com/docs/10.x/redirects#redirecting-controller-actions
With that being said, I've since explored a bit more of the available options and decided that the better approach would be following the steps I explained in the Slack thread, since it incorporates our needs in the already established code from Fortify and Jetstream. What do you think? I still have to test it and verify that it does work, but I currently don't see where it could go wrong (famous last words).

pedrosilva17:
CV support is pratically done. I tried to utilize the already established Fortify and Jetstream classes/methods as much as possible and expand them where needed. 
In my last commit, some temporary changes to schemas were made in order to fix an ongoing problem (being unable to create new accounts), but shouldn't be considered valid solutions and will be reverted when merging. In that same commit, I modified the default frontend to enable CV upload for testing, so please feel free to do so.
Currently there's only one route for downloads, but that could change, since we might need to specify different storage disks depending on the type of file we're downloading (the name of the storage disk is hardcoded in the test frontend, at the moment).
As usual, leave your suggestions/fixes below, I still believe there could be room for improvement.

app/Traits/HasCV.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * Get the disk that CVs should be stored on.
+     *
+     * @return string
+     */
+    protected function CVDisk()
+    {
+        return isset($_ENV['VAPOR_ARTIFACT_NAME']) ? 's3' : config('jetstream.cv_disk', 'public');
```
Naapperas: We are neither using Vapor or AWS S3 buckets, should this be here?

routes/jetstream.php
```diff
@@ -38,6 +39,9 @@
 
         Route::delete('/user/profile-photo', [ProfilePhotoController::class, 'destroy'])
             ->name('current-user-photo.destroy');
+            
+        Route::delete('/user/cv', [CVController::class, 'destroy'])
```
Naapperas: I believe we should have a route group for the CV. Also, the `user` routes could make use of route parameters, maybe I'm jumping to conclusions but I see none.

routes/api.php
```diff
@@ -17,3 +18,5 @@
 Route::middleware('auth:sanctum')->get('/user', function (Request $request) {
     return $request->user();
 });
+
+Route::get('/download', [FileController::class, 'download'])->name('file.download');
```
Naapperas: I believe that downloads can already be handled by the browser, is this needed?

Naapperas (changes requested):

app/Traits/HasCV.php
pedrosilva17: Could be useful in case we add remote storage capabilities, isn't needed for now though

routes/jetstream.php
pedrosilva17: These are the default jetstream routes (excluding the CV part). But yes, we would need to pass parameters to the routes if we want to access other user profiles

routes/api.php
pedrosilva17: I use that route to generate the response for downloading the file in the browser (with Laravel's `Storage::download()`). Is there a cleaner way of generating that response, passing the file storage url, using only the browser?

app/Traits/HasCV.php
Naapperas: Oh ok, like some cloud provider. Forgot that, nice catch. The only catch I can ser with this approach (which could be easily fixable) os if the provider we choose has a custom PHP SDK, in which case these would not be needed.

routes/jetstream.php
Naapperas: Perhaps this needs a dedicated discussion just on this topic, Will leave for later.

routes/api.php
Naapperas: I think that browsers can already handle URI's that refer to files, either for downloading or displaying, but I'm not entirely certain. We need to investigate further.

database/migrations/2023_07_04_010134_create_schema.php
```diff
@@ -24,9 +24,6 @@ public function up(): void {
         });
 
         // User tables
```
CiscoPr: Everything lgtm, but i have a question: won't there be any problem by eliminating the **users** table that was here?

pedrosilva17: I completely forgot to add that back after testing, my bad

pedrosilva17:
Now sure what's up with the tests - some easy ones fail, like rendering the registration page or creating a new account, but I can do that just fine in my local branch.

database/migrations/2014_10_12_000000_create_users_table.php
```diff
@@ -13,13 +13,15 @@ public function up(): void
...
             $table->id();
+            $table->morphs('usertype');
             $table->string('name');
             $table->string('email')->unique();
             $table->timestamp('email_verified_at')->nullable();
             $table->string('password');
             $table->rememberToken();
             $table->foreignId('current_team_id')->nullable();
             $table->string('profile_photo_path', 2048)->nullable();
+            $table->string('cv_path', 2048)->nullable();
```
ttoino: Migrations should not be modified after they are merged, this is because the website could already be in production and we would not be able to run all migrations from the start, this should be separated into a new migration.

database/migrations/2023_07_04_010134_create_schema.php
```diff
@@ -26,9 +26,6 @@ public function up(): void
...
-        Schema::table('users', function (Blueprint $table) {
-            $table->morphs('usertype');
-        });
```
ttoino: Same as above

database/migrations/2023_08_04_183516_create_resumes_table.php
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('resumes', function (Blueprint $table) {
+            $table->id();
+            $table->text('resume');
+            $table->timestamps();
+        });
```
ttoino: Why is a separate table required for the resumes? Isn't it already an attribute in the user model?

app/Providers/JetstreamServiceProvider.php
```diff
@@ -13,7 +13,7 @@ class JetstreamServiceProvider extends ServiceProvider
      */
     public function register(): void
     {
-        //
+        Jetstream::ignoreRoutes();
```
ttoino, line 17: What does this do?

app/Providers/FortifyServiceProvider.php
```diff
@@ -19,7 +19,7 @@ class FortifyServiceProvider extends ServiceProvider
      */
     public function register(): void
     {
-        //
+        Fortify::ignoreRoutes();
```
ttoino, line 23: Same as above

.env
```diff
@@ -2,7 +2,7 @@ APP_NAME="SINF Website 2023"
 APP_ENV=local # Use "maintenance" to enable the maitenance page
 APP_KEY=base64:AVegBkyjHs2QQsx53M9Kt9GbnrDLc6yG4cRg+WMJrbA=
 APP_DEBUG=true
-APP_URL=http://localhost
+#APP_URL=http://localhost
```
ttoino: Why was this commented?

routes/api.php
ttoino: This route is not protected, this could lead anyone to another user's cv I think

ttoino (changes requested):

app/Providers/JetstreamServiceProvider.php
ttoino: Nevermind, I got it

database/migrations/2014_10_12_000000_create_users_table.php
pedrosilva17: You're right, my mistake

database/migrations/2023_08_04_183516_create_resumes_table.php
pedrosilva17: This is outdated code from my first attempt at implementing file uploads. I'll clean it up when I return

.env
pedrosilva17: No clue, honestly, I'll change it back

ttoino:
Any updates on this PR? Work here will unlock work on other issues where user routes are needed, such as #41 and #43

pedrosilva17:
I will not be home until 7th September at night, so I can't work on this at the moment. If it is urgent, someone else can apply the required changes, otherwise I will do it as soon as possible.

Naapperas:
@pedrosilva17 What's missing in this PR? The changes you made here are important to other issues so it's important to have this one merged as soon as possible.

pedrosilva17:
@CiscoPr has suggested we make these changes (the ones in regards to CVs) in the Student table instead of the User, which would require more work but would make the interactions with Companies and Students easier when joining stuff (regarding visits, CVs, etc.). What do you think @Naapperas ?

Naapperas:
> @CiscoPr has suggested we make these changes (the ones in regards to CVs) in the Student table instead of the User, which would require more work but would make the interactions with Companies and Students easier when joining stuff (regarding visits, CVs, etc.). What do you think @Naapperas ?

Sure, go ahead.

database/migrations/2014_10_12_000000_create_users_table.php
```diff
@@ -13,6 +13,7 @@ public function up(): void
     {
         Schema::create('users', function (Blueprint $table) {
             $table->id();
+            $table->morphs('usertype');
```
ttoino: This migration should not be modified

pedrosilva17: Forgot to revert that change, thanks 👍

app/Actions/Fortify/UpdateUserProfileInformation.php
```diff
@@ -21,12 +22,18 @@ public function update(User $user, array $input): void
             'name' => ['required', 'string', 'max:255'],
             'email' => ['required', 'email', 'max:255', Rule::unique('users')->ignore($user->id)],
             'photo' => ['nullable', 'mimes:jpg,jpeg,png', 'max:1024'],
+            'cv' => ['nullable', 'mimes:pdf', 'max:4096'],
```
Naapperas: Please take a look at the documentation [here](https://laravel.com/docs/10.x/validation#validating-files) and see what is the best way to define these rules.

app/Actions/Fortify/UpdateUserProfileInformation.php
```diff
@@ -21,12 +22,18 @@ public function update(User $user, array $input): void
...
         ])->validateWithBag('updateProfileInformation');
 
         if (isset($input['photo'])) {
             $user->updateProfilePhoto($input['photo']);
         }
 
+        if (isset($input['cv'])) {
+            $student = Student::where('user_id', $user->id)->first();
```
Naapperas:
```suggestion
            $student = Student::find($user->id);
```

This uses the primary key directly, whatever it may be. I believe it's a better way of doing this but feel free to ignore.

app/Traits/HasCV.php
Naapperas: @pedrosilva17 we will be storing images and files on our hardware so we won't need any cloud provider,, I'll leave it up to you to change this line (or not) and mark this conversation as resolved.

app/Models/Student.php
```diff
@@ -21,6 +23,20 @@ class Student extends Model
...
+     * @var array<int, string>
+     */
+    protected $appends = [
+        'cv_url',
+    ];
+
+    public function getCvUrlAttribute()
+    {
+        return $this->cv_path;
+    }
```
Naapperas: This method is not used anywhere, please remove it.

app/Traits/HasCV.php
```diff
@@ -0,0 +1,71 @@
...
+{
+    /**
+     * Update the student's CV.
+     *
+     * @param  string  $storagePath
+     * @return void
+     */
+    public function updateCV(UploadedFile $cv, $storagePath = 'cvs')
+    {
+        tap($this->cv_path, function ($previous) use ($cv, $storagePath) {
```
Naapperas, line 19: Why did you use `tap()` here?

app/Traits/HasCV.php
```diff
@@ -0,0 +1,71 @@
...
+            $this->forceFill([
+                'cv_path' => $cv->storePublicly(
+                    $storagePath, ['disk' => $this->CVDisk()]
+                ),
+            ])->save();
+
+            if ($previous) {
+                Storage::disk($this->CVDisk())->delete($previous);
+            }
```
Naapperas: This should not be here: semantically an update is not a creation action.

Naapperas: Eitherway, I believe that this, if it is to stay in this method, should happen before we store the new file.

app/Traits/HasCV.php
```diff
@@ -0,0 +1,71 @@
...
+     *
+     * @param  string  $storagePath
+     * @return void
+     */
+    public function updateCV(UploadedFile $cv, $storagePath = 'cvs')
+    {
+        tap($this->cv_path, function ($previous) use ($cv, $storagePath) {
+            $this->forceFill([
+                'cv_path' => $cv->storePublicly(
+                    $storagePath, ['disk' => $this->CVDisk()]
```
Naapperas, line 26: The documentation I found did not speak of having an array as an argument, does this work as it should?

app/Traits/HasCV.php
```diff
@@ -0,0 +1,71 @@
...
+namespace App\Traits;
+
+use Illuminate\Database\Eloquent\Casts\Attribute;
+use Illuminate\Http\UploadedFile;
+use Illuminate\Support\Facades\Storage;
+
+trait HasCV
+{
+    /**
+     * Update the student's CV.
```
Naapperas: In this case it is the students' CVs, but I believe this comment should be more generic.

routes/api.php
```diff
@@ -17,3 +18,5 @@
 Route::middleware('auth:sanctum')->get('/user', function (Request $request) {
     return $request->user();
 });
+
+Route::middleware('auth:sanctum')->get('/download', [FileController::class, 'download'])->name('file.download');
```
Naapperas: Why are we doing this through an API call? API support isn't enabled yet so this should be based on the web router.

routes/jetstream.php
Naapperas: Coming back to this comment. Even though we are manually adding routes in order to have control over the route paths, we should not change these files any further. We should use the `web.php` file for any web related route.

app/Models/Resume.php
Naapperas, line 1: This class is not used anywhere, it should be removed.

routes/api.php
Naapperas: In my opinion, we should have a route that is like `/user/cv/download` and that is the one that triggers the download.

routes/jetstream.php
Naapperas: Also, companies need to be able to see users that have visited them, so we should still have a route group for that.

app/Http/Controllers/FileController.php
```diff
@@ -0,0 +1,16 @@
...
+use App\Traits\ProcessesFiles;
+use Illuminate\Http\Request;
+
+class FileController extends Controller
+{
+    use ProcessesFiles;
+
+    public function download(Request $request)
+    {
+        return $this->downloadFromUrl($request->url, $request->storageDisk);
```
Naapperas: I think we should assume for now that the storage disk will always the `public` disk. If not, we should use the configuration used for storing them, instead of relying on the request. Food for thought.

app/Traits/ProcessesFiles.php
```diff
@@ -0,0 +1,13 @@
...
+
+namespace App\Traits;
+
+use Illuminate\Support\Facades\Storage;
+
+trait ProcessesFiles
+{
+    public function downloadFromUrl($url, $storageName)
+    {
+        return Storage::disk($storageName)->download($url);
```
Naapperas: Same as above.

app/Http/Controllers/CVController.php
Naapperas, line 1: I'm not really fond of having a whole controller just to delete CVs, even if that is what Jetstream is doing.

Naapperas (changes requested):
I left a lot of comments on this one.

Perhaps we should drop the whole "generic file upload" thing and just focus on the CVs for now. I believe it would not be that big of a change to make, simply deleting the `FileController` and moving some methods and routes around. This way we can close this PR for good and advance the site's backend.

Comments? @ttoino @toni-santos @pedrosilva17 @CiscoPr @diogofonte @pedronunes19

Naapperas:
Also, if it is needed just ignore the tests. Other than that, some cleaning should be made to the commits.

Naapperas:
I'm at a loss regarding the tests. Unless someone figures this out we are probably going to have to ignore them.

app/Traits/HasCV.php
pedrosilva17: At first I thought it would be more useful, but all it does is avoid writing `$previous = $this->cv_path` in a separate line.

app/Traits/HasCV.php
pedrosilva17: This was adapted directly from the `HasProfilePhoto.php` file in Jetstream, which deals with profile pictures.

Naapperas:
And don't forget to rebase with `main`.

routes/jetstream.php
Naapperas: This is done in #48

routes/web.php
```diff
@@ -32,4 +32,6 @@
     Route::get('/dashboard', function () {
         return Inertia::render('Dashboard');
     })->name('dashboard');
+    Route::delete('/user/cv', [CVController::class, 'destroy'])->name('current-user-cv.destroy');
+    Route::get('/user/cv/download', [FileController::class, 'download'])->name('file.download');
```
Naapperas: We should make use of [route groups](https://laravel.com/docs/10.x/routing#route-groups), makes the code cleaner: I suggest a group for the users and a group for CVs.

app/Actions/Jetstream/DeleteUser.php
```diff
@@ -12,6 +13,10 @@ class DeleteUser implements DeletesUsers
      */
     public function delete(User $user): void
     {
+        $student = Student::where('user_id', $user->id)->first();
+        if ($student) {
+            $student->deleteCV();
+        }
```
Naapperas: I believe it would be best if the model came from the `morph` relationship instead of calling `Student::where`.

Naapperas (changes requested):
Please leave a comment in the Jestream and Fortify files (and Providers aswell) explaining why we have that code in place.

Naapperas (dismissed):

Naapperas:
Don't forget to rebase

Naapperas:
> Don't forget to rebase

By doing this you can also tackle the comment that's left.

app/Actions/Fortify/UpdateUserProfileInformation.php
```diff
@@ -21,12 +23,18 @@ public function update(User $user, array $input): void
             'name' => ['required', 'string', 'max:255'],
             'email' => ['required', 'email', 'max:255', Rule::unique('users')->ignore($user->id)],
             'photo' => ['nullable', 'mimes:jpg,jpeg,png', 'max:1024'],
+            'cv' => ['nullable', File::types(['pdf'])->max(4 * 1024)],
```
ttoino: Is this maximum size in bytes or another unit? If it's in bytes I'm not sure that's enough, as my CV takes up about 500KB

app/Actions/Jetstream/DeleteUser.php
ttoino: These docs may be helpful here: https://laravel.com/docs/10.x/eloquent-relationships#querying-morph-to-relationships

ttoino (dismissed):

app/Actions/Fortify/UpdateUserProfileInformation.php
Naapperas: @pedrosilva17 according to [this](https://laravel.com/docs/10.x/validation#validating-files-file-sizes) you can specify the units directly to the validator. Makes it more readable.

Naapperas:
Don't forget to rebase

ttoino (approved):

Naapperas (approved):

PR #22, opened 2023-08-08
Student history

### LINKED ISSUE #15: Student visit history
Companies that are [gold or higher](https://github.com/semanadeinformatica/website-2023/issues/14) need to have access to a log of the students that have paid them a visit, along with these student's CVs.

CiscoPr:
Closes #15 
A separate migration updates the new table **visit_history**  based on the values from the newly inserted row in the **enrollment_quest** table, including the corresponding **student_id**, **company_id**, and **timestamps**

### DIFF
--- a/database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
+++ b/database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
@@ -1,0 +1,75 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        // Visit history
+        Schema::create('visit_history', function (Blueprint $table) {
+            $table->id();
+            $table->foreignIdFor(\App\Models\Student::class)->constrained()->cascadeOnDelete();
+            $table->foreignIdFor(\App\Models\Company::class)->constrained()->cascadeOnDelete();
+            $table->timestamps();
+        });
+
+        //Triggers
+        DB::unprepared('
+            CREATE OR REPLACE FUNCTION update_visit_history() RETURNS TRIGGER AS $$
+                BEGIN
+                    DECLARE v_company_tier TEXT;
+                    DECLARE v_company_id INT;
+                    BEGIN
+                        IF (NEW IS NOT NULL) THEN
+
+                            SELECT c.tier
+                            INTO v_company_tier
+                            FROM quests q
+                            LEFT JOIN companies c ON q.requirement_type = \'\App\Models\Company\' AND q.requirement_id = c.id
+                            WHERE q.id = NEW.quest_id;
+
+                            IF ((v_company_tier != \'SILVER\') AND (SELECT requirement_type FROM quests WHERE id = NEW.quest_id) = \'\App\Models\Company\')THEN
+
+                                SELECT requirement_id INTO v_company_id FROM quests WHERE id = NEW.quest_id;
+
+                                INSERT INTO visit_history (student_id, company_id, created_at, updated_at)
+                                SELECT
+                                    e.student_id,
+                                    v_company_id,
+                                    NOW(),
+                                    NOW()
+                                FROM enrollments e
+                                LEFT JOIN students s ON e.student_id = s.id
+                                LEFT JOIN users u ON s.user_id = u.id
+                                WHERE e.id = NEW.enrollment_id;
+                            END IF;
+                        END IF;
+                    END;
+                    RETURN NEW;
+                END;
+            $$ LANGUAGE plpgsql;
+
+    ');
+
+        DB::unprepared('
+            CREATE TRIGGER update_visit_history
+            AFTER INSERT ON enrollment_quest
+            FOR EACH ROW EXECUTE FUNCTION update_visit_history();
+        ');
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        Schema::dropIfExists('update_visit_history');
+        Schema::dropIfExists('visit_history');
+    }
+};


diogofonte (dismissed):
Looks good to me!

database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
```diff
@@ -0,0 +1,52 @@
...
+            $table->timestamps();
+        });
+
+        // Triggers
+        DB::unprepared('
+            CREATE OR REPLACE FUNCTION update_visit_history() RETURNS TRIGGER AS $$
+                BEGIN
+                    IF (NEW IS NOT NULL) THEN
+                        INSERT INTO visit_history (student_id, company_id, created_at, updated_at)
+                        VALUES (NEW.student_id, (
```
pedronunes19: doesn't enrollment_quest have an enrollment_id instead? the overall solution seems to work, that detail just propped some questions

pedronunes19 (changes requested):
This seems to achieve our goal, not sure if anything else is needed in the context of this PR

database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
CiscoPr: Yes, you're absolutely right, that detail didn't even cross my mind. I will correct it asap

CiscoPr:
Now, although the student_id is finally selected correctly, i came across another mistake on my part when it comes to the company_id, which is not properly retrieved

CiscoPr:
The **company_id** should also be correctly retrieved now. The previous solution was wrong, since the id of the quests table did **not** correspond to the id's of the companies. 
Right now, what I do is to work with the **polymorphic relationship** "requirement" of the quests table and, if it points to the **Company** model, I select its id.

pedronunes19 (dismissed):
LGTM

pedronunes19 (dismissed):

database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
```diff
@@ -0,0 +1,74 @@
...
+        ');
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        Schema::dropIfExists('visit_history');
+    }
```
ttoino: Is this enough? Is dropping the trigger not required?

CiscoPr: i don't think so. I think the trigger will be constantly updated everytime the migration is run. Not 100% sure tho

ttoino: I checked our other migration, and the function `update_visit_history` should also be dropped here. It's not that bad if it isn't, as you're creating it only if it doesn't exist, but all migrations should be self contained and completely reversible.

CiscoPr: Makes sense. I will change it then, thanks for your feedback

ttoino (dismissed):

pedronunes19 (dismissed):

CiscoPr:
I added the cv_path to the trigger, even tho this will only fully make sense when the PR corresponding to the [file uploading](https://github.com/semanadeinformatica/website-2023/pull/16) is merged

CiscoPr:
Right now, all checks are passing, but, for some reason, while running the tests, there's the following error:
`  SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "IF"
LINE 6:                     IF (NEW IS NOT NULL) THEN`

CiscoPr:
This problem should now be solved

database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
```diff
@@ -0,0 +1,80 @@
...
+                        IF (NEW IS NOT NULL) THEN
+
+                            SELECT c.tier
+                            INTO v_company_tier
+                            FROM quests q
+                            LEFT JOIN companies c ON q.requirement_type = \'\App\Models\Company\' AND q.requirement_id = c.id
+                            WHERE q.id = NEW.quest_id;
+
+                            IF v_company_tier != \'SILVER\' THEN
+                                INSERT INTO visit_history (student_id, company_id, cv_path, created_at, updated_at)
```
ttoino: Why do you insert a `cv_path` here when there is no such attribute in the table definition?
Maybe this was forgotten but this will give rise to issues in the future when this trigger runs this function.

database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
```diff
@@ -0,0 +1,80 @@
...
+                                SELECT
+                                    e.student_id,
+                                    (
+                                        CASE
+                                            WHEN (
+                                                SELECT requirement_type FROM quests WHERE id = NEW.quest_id
+                                            ) = \'\App\Models\Company\' THEN
+                                                (SELECT requirement_id FROM quests WHERE id = NEW.quest_id)
+                                            ELSE NULL
```
ttoino: Since you are reusing the same value here it could be extracted to a variable.

database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
```diff
@@ -0,0 +1,80 @@
...
+                                        END
+                                    ),
+                                    u.cv_path,
```
ttoino: `cv_path` belongs to the Student, not the User

database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
ttoino: Inserting `NULL` here will cause errors since you defined the foreign key to be constrained by the DB (aka it cannot be `NULL`).

One solution is to see if the requirement type refers to a company and only then you insert anything.

database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
```diff
@@ -0,0 +1,80 @@
...
+                        END IF;
+                    END;
+                    RETURN NEW;
+                END;
+            $$ LANGUAGE plpgsql;
+
+    ');
+
+        DB::unprepared('
+            CREATE TRIGGER update_visit_history
```
ttoino, line 61: Just for safety, I believe we should include a `IF NOT EXISTS` clause when creating the trigger.

database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
```diff
@@ -0,0 +1,80 @@
...
+            CREATE OR REPLACE FUNCTION update_visit_history() RETURNS TRIGGER AS $$
+                BEGIN
+                    DECLARE v_company_tier TEXT;
+                    BEGIN
+                        IF (NEW IS NOT NULL) THEN
+
+                            SELECT c.tier
+                            INTO v_company_tier
+                            FROM quests q
+                            LEFT JOIN companies c ON q.requirement_type = \'\App\Models\Company\' AND q.requirement_id = c.id
```
ttoino, line 34: You get the tier for a company because the requirement type specifies that this quest referred to a company, but then you re(dundantly)-check the requirement type again.

database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
ttoino: First confirm if this is even possible.

ttoino (changes requested):
I've left some comments on the trigger function, if it works then it is good for now but I believe it could/should be improved in order to not have slow running code on the DB: I thought this up in a hurry but I believe you can implement this using only one insert, more `JOIN`s and more complex conditions (for requirement types and company tiers).

Ass: @Naapperas

database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
CiscoPr: Yes, i only noticed this mistake on my part during the group meeting on monday, sorry about that

database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
CiscoPr: Yes, ideally that's the case. I just got confused on the PR #16 so i put in the user.
Also, @toni-santos and @pedronunes19 mentioned something important: since we already have the `student_id` in here (and by correcting the usage of the `cv_path` in the **File Upload** PR in order for it to belong to the student instead of the user) adding the column of the cv becomes a bit useless, don't you think so? The reason being that, by having the student id, the company will already have access to the cv of the student

database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
CiscoPr: ~~could you be more specific please?~~
nvm this, i think i got it

Naapperas:
@CiscoPr I have opened a [PR](https://github.com/semanadeinformatica/website-2023/pull/50) that may affect your trigger function, beware of that.

CiscoPr:
> @CiscoPr I have opened a [PR](https://github.com/semanadeinformatica/website-2023/pull/50) that may affect your trigger function, beware of that.

Okay, i will check it out then

CiscoPr:
@Naapperas Just to make sure i understood correctly the sugestion you made [here](https://github.com/semanadeinformatica/website-2023/pull/22#discussion_r1316395289), is this what you were talking about? 
```
DECLARE q_ids TEXT;
DECLARE company_type TEXT;
```
```
q_ids := \'FROM quests WHERE id = NEW.quest_id\';
company_type := \'\App\Models\Company\';
```
```
SELECT requirement_type q_ids ) = company_type THEN (SELECT requirement_id q_ids)
```
Btw, these changes still don't take care of [this](https://github.com/semanadeinformatica/website-2023/pull/22#discussion_r1316396952) problem... i will look into it now

database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
CiscoPr: I searched a bit and tested it out, but it seems like it is impossible to do this. I tested it out and it even prints out a ```SYNTAX ERROR```

Naapperas: `CREATE OR REPLACE` is what I was looking for. It is not that important thought, hoorah migrations.

database/migrations/2023_08_05_121302_add_oncreate_trigger_enrollment_quest.php
```diff
@@ -0,0 +1,80 @@
...
+                                    (SELECT requirement_type FROM quests WHERE id = NEW.quest_id) = \'\App\Models\Company\'
+                                    )
+                                THEN 
+                                    SELECT requirement_id INTO v_company_id FROM quests WHERE id = NEW.quest_id;
+                                END IF;
+
+                                INSERT INTO visit_history (student_id, company_id, created_at, updated_at)
+                                SELECT
+                                    e.student_id,
+                                    v_company_id,
```
Naapperas: This suffers from the same problem I mentioned: if the requirement is not met, you are inserting the default value of `v_company_id` (which I assume is `NULL`) into the table, leading to the same issue I talked about. What you should do is a check similar to what you do for the company tier.

Naapperas (changes requested):
Almost there.

ttoino (approved):

pedronunes19 (approved):

Naapperas (approved):
LGTM! Amazing work @CiscoPr  :two_hearts: :two_hearts::two_hearts:  

Thank very much you @ttoino and @pedronunes19 for reviewing this PR :heart_eyes:

PR #23, opened 2023-08-08
Auth pages

ttoino:
Implements and styles the authentication pages (login, register, recover password, etc.)


resources/js/Components/Checkbox.vue
```diff
@@ -31,6 +31,6 @@ const proxyChecked = computed({
         v-model="proxyChecked"
         type="checkbox"
         :value="value"
-        class="rounded dark:bg-gray-900 border-gray-300 dark:border-gray-700 text-indigo-600 shadow-sm focus:ring-indigo-500 dark:focus:ring-indigo-600 dark:focus:ring-offset-gray-800"
+        class="border border-black checked:border-black checked:hover:border-black bg-2023-bg hover:shadow shadow-2023-orange checked:bg-2023-teal checked:hover:bg-2023-teal-dark outline-none transition-all"
```
Naapperas, line 34: Aren't some of these classes redundant? Namely the ones referring to the border.

resources/js/Pages/Auth/Login.vue
```diff
@@ -30,61 +29,49 @@ const submit = () => {
...
 
-            <div class="block mt-4">
-                <label class="flex items-center">
-                    <Checkbox v-model:checked="form.remember" name="remember" />
-                    <span class="ml-2 text-sm text-gray-600 dark:text-gray-400">Remember me</span>
-                </label>
-            </div>
+            <label class="flex items-center gap-2 self-stretch">
+                <Checkbox v-model:checked="form.remember" name="remember" />
+                <span class="text-2023-teal">Remember me</span>
```
Naapperas: This text should be in Portuguese, to match the rest of the page.

tailwind.config.js
```diff
@@ -17,8 +19,41 @@ export default {
...
                 sans: ['Figtree', ...defaultTheme.fontFamily.sans],
             },
+            colors: {
+                '2023-bg': 'rgb(248, 245, 231)',
+                '2023-red': 'rgb(217, 79, 4)',
+                '2023-red-dark': 'rgb(177, 93, 93)',
+                '2023-teal': 'rgb(0, 113, 114)',
+                '2023-teal-dark': 'rgb(2, 82, 89)',
+                '2023-orange': 'rgb(242, 147, 37)',
+            },
```
Naapperas, line 29: The color palette should be configurable outside of this file, so that future developers have an easier experience adapting their version of the website.

tailwind.config.js
```diff
@@ -17,8 +19,41 @@ export default {
...
+                    'box-shadow': [
+                        `var(--tw-ring-offset-shadow, 0 0 #0000)`,
+                        `var(--tw-ring-shadow, 0 0 #0000)`,
+                        `var(--tw-shadow)`,
+                    ].join(', '),
+                }
+            },
+        },
+        { values: theme('boxShadow'), type: 'shadow' })
+    })],
```
Naapperas, line 58: I don't know if I'm too confortable with these options being hard-coded for the same reason I expressed above: they may make sense for the design of this year's edition but not for the next one. I will write a separate issue for that but please keep that in mind.

tailwind.config.js
Naapperas: I believe this can already be updated on this PR, since it's a relatively small fix.

Naapperas (changes requested):

resources/js/Components/Checkbox.vue
ttoino: No, `border-black` only sets the border color, while `border` sets the border to be 1px wide and gray. The ones with variants are needed to override the default styles set by tailwind-forms

Naapperas: Oh ok, kinda sucks to have the same class name repeated but it is what it is.

tailwind.config.js
ttoino: In an effort to keep this pr small, I think this should be explored in a separate issue that tries to make all the styles generic enough to be different for different editions.

Naapperas (approved):

pedronunes19 (approved):

PR #25, opened 2023-08-08
Maintenance page

### LINKED ISSUE #24: Maintenance page
Create a maintenance page for the eventuality of service down times and as a placeholder for the upcoming event.

toni-santos:
Closes #24


app/Exceptions/Handler.php
```diff
@@ -4,9 +4,38 @@
...
+     * Prepare exception for rendering.
+     *
+     * @param  \Throwable  $e
+     * @return \Throwable
+     */
+    public function render($request, Throwable $e)
+    {
+        $response = parent::render($request, $e);
+        // ! app()->environment(['local', 'testing']) &&
+        if (in_array($response->status(), [500, 404, 403])) {
```
Naapperas: Why is this array special? What is the meaning of those error codes?

Naapperas: All this to say, try not no use too many magic constants :)

app/Exceptions/Handler.php
```diff
@@ -4,9 +4,38 @@
...
+                ->setStatusCode($response->status());
+        } elseif ($response->status() === 503) {
+            return Inertia::render('Maintenance')
+                ->toResponse($request)
+                ->setStatusCode($response->status());
+        } elseif ($response->status() === 419) {
+            return back()->with([
+                'message' => 'The page expired, please try again.',
+            ]);
+        }
```
Naapperas, line 37: Is this here because we are overriding the default handler? One would think that this could be done automatically. Also, are there other ways of handling error requests? Like appending an additional handler to a "stack" of handler? This way we could preserve this file and do things in a more `Laravel`-ish way. If there is no other way then that's fine.

resources/js/Pages/Error.vue
```diff
@@ -0,0 +1,28 @@
...
+  }[props.status]
+})
+
+const description = computed(() => {
+  return {
+    500: 'Whoops, something went wrong on our servers.',
+    404: 'Sorry, the page you are looking for could not be found.',
+    403: 'Sorry, you are forbidden from accessing this page.',
+  }[props.status]
+})
```
Naapperas, line 20: These could be made into one single `computed` call that returns an object. I don't know if that makes a difference tho, if it does not then ignore this comment.

resources/views/errors/minimal.blade.php
```diff
@@ -0,0 +1,34 @@
...
...2,1)}}@media (min-width:640px){.sm\:rounded-lg{border-radius:.5rem}.sm\:block{display:block}.sm\:items-center{align-items:center}.sm\:justify-start{justify-content:flex-start}.sm\:justify-between{justify-content:space-between}.sm\:h-20{height:5rem}.sm\:ml-0{margin-left:0}.sm\:px-6{padding-left:1.5rem;padding-right:1.5rem}.sm\:pt-0{padding-top:0}.sm\:text-left{text-align:left}.sm\:text-right{text-align:right}}@media (min-width:768px){.md\:border-t-0{border-top-width:0}.md\:border-l{border-left-width:1px}.md\:grid-cols-2{grid-template-columns:repeat(2,minmax(0,1fr))}}@media (min-width:1024px){.lg\:px-8{padding-left:2rem;padding-right:2rem}}@media (prefers-color-scheme:dark){.dark\:bg-gray-800{--bg-opacity:1;background-color:#2d3748;background-color:rgba(45,55,72,var(--bg-opacity))}.dark\:bg-gray-900{--bg-opacity:1;background-color:#1a202c;background-color:rgba(26,32,44,var(--bg-opacity))}.dark\:border-gray-700{--border-opacity:1;border-color:#4a5568;border-color:rgba(74,85,104,var(--border-opacity))}.dark\:text-white{--text-opacity:1;color:#fff;color:rgba(255,255,255,var(--text-opacity))}.dark\:text-gray-400{--text-opacity:1;color:#cbd5e0;color:rgba(203,213,224,var(--text-opacity))}}
```
Naapperas, line 10: Where did you take this from?

Naapperas (changes requested):

resources/views/errors/minimal.blade.php
toni-santos: That was generated by laravel using `php artisan vendor:publish ` and the `laravel-errors` option.

app/Exceptions/Handler.php
toni-santos: This was directly taken from [Inertia's documentation](https://inertiajs.com/error-handling#production), furthermore, as you can see by the commented line, it has been altered to make testing in this branch easier (we need to evidently revert it when pushing to main).

app/Exceptions/Handler.php
toni-santos: Like before, [Inertia's documentation](https://inertiajs.com/error-handling#production) is the source. As far as making it in a more `Laravel`-ish way, I'm not sure, I can look into it once we have the remaining, more pressing issues tackled :)

app/Exceptions/Handler.php
Naapperas: Understood, but then it's their fault for having their code like that. We can do better and Pull those values into named constants :)

This allows for easier updates in the future, should they be needed.

Regarding the commented line, we could have something like a "feature flag", like "doNotTestDevAsProd" which, when or'ed with the commented expression, could impose a programmatic toggle on that condition. Maybe I'm complicating things but then, when we are ready to deploy the website, it's just a matter of changing that same feature flag (which could be kept as an environment variable).

What do you think?

app/Exceptions/Handler.php
Naapperas: Ok, sounds good, leaving this conversation open so it's easier to track.

resources/views/errors/minimal.blade.php
Naapperas: Oh, ok. Kinda weird for them to have all those styles like that but then that is that. Maybe que should clean them up but I'll leave that completely up to you.

app/Exceptions/Handler.php
toni-santos: From what I read regarding this code, it seems that it is already controlled by an environmental flag, that can be `production`, `local` or `testing`, we could simply add a state to this flag like `maintenance_test` that would overwrite the other ones in the if condition. I think that's similar to what you are proposing, no?

resources/views/errors/minimal.blade.php
toni-santos: I think that's a concern to another issue/PR since it concerns error pages rather than the maintenance page. Regardless, we are not using any of these templates since we are using Inertia for maintenance/errors.

app/Exceptions/Handler.php
Naapperas: It is yes. It all boils down to a matter of taste and consistency.

toni-santos: In order to make this easier, and to avoid accidentally making the site go into maintenance mode I suggest adding `maintenance` to the environments. The program flow would go as follows:
```php
        $DEVELOPMENT = ['maintenance', 'local', 'testing'];
        $ERROR_PAGES = [500, 404, 403];
        
        if (!app()->environment($DEVELOPMENT) && in_array($response->status(), $ERROR_PAGES)) {
            return Inertia::render('Error', ['status' => $response->status()])
                ->toResponse($request)
                ->setStatusCode($response->status());
        } elseif ($response->status() === 503 && app()->environment(['maintenance'])) {
            return Inertia::render('Maintenance')
                ->toResponse($request)
                ->setStatusCode($response->status());
        } elseif ($response->status() === 419) {
            return back()->with([
                'message' => 'The page expired, please try again.',
            ]);
        }
```

What do you think of this solution?

toni-santos: I forgot to mention but to even make Laravel go into maintenance mode, `php artisan down` must be ran (and ` php artisan up` in order to get it back up). This solution also allows us to have a custom 503 page for regular unavailability errors.

toni-santos:
This is ready for a thorough review and merge.

pedrosilva17 (dismissed):
🚀🚀

pedronunes19 (dismissed):

.env
```diff
@@ -1,5 +1,5 @@
 APP_NAME="SINF Website 2023"
-APP_ENV=local
+APP_ENV=local # Use maintenance_test to test the maitenance page
```
Naapperas: You use `maintenance` in the code, change this accordingly

resources/js/Pages/Error.vue
Naapperas: @toni-santos any thoughts on this?

resources/js/Pages/Maintenance.vue
```diff
@@ -0,0 +1,73 @@
+<script setup>
```
Naapperas, line 1: Do the paths on the images created here always work? Pure curiosity, this is something I really don't know.

Naapperas (changes requested):

resources/js/Pages/Error.vue
toni-santos: I don't believe it would make a difference, regardless, I believe this is an issue to be tackled at another time, seeing as it relates solely to error pages.

resources/js/Pages/Maintenance.vue
toni-santos: From my testing and my cursory investigation it seems they do, but now that you mention I see that the images have not been committed. I believe they should be present in the `public` folder, whereas I placed them inside `storage`, I'll take a look at it.

Naapperas (approved):

pedronunes19 (approved):

PR #31, opened 2023-08-19
Setup CI stack

### LINKED ISSUE #28: Add checks
We should exploit Github's features and add CI/CD to PR's. Needs to be further investigated.

ttoino:
Adds and configures workflows to test the application and check the quality of the code.

Adds and sets up the following tools:

- Laravel Pint (for php formatting)
- Prettier (for javascript formatting)
- Eslint (for javascript linting)
- Typescript and vue-tsc (for javascript type checking)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,112 @@
...
+    name: Run tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+ 
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.2'
```
Naapperas: This version should come from a matrix or an externally defined variável, in my opinion.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,112 @@
...
+          extensions: dom, curl, libxml, mbstring, zip, pcntl, pdo, sqlite, pdo_sqlite, bcmath, soap, intl, gd, exif, iconv
```
Naapperas: Same thing here, if possible

.github/workflows/ci.yml
```diff
@@ -0,0 +1,112 @@
...
+          coverage: none
+ 
+      - name: Run composer install
+        run: composer install -n --prefer-dist
+ 
+      - name: Run Pint
+        run: ./vendor/bin/pint --test
+
+  prettier:
+    name: Check Typescript code style
```
Naapperas, line 76: Shouldn't this be "Format..."?

Naapperas (changes requested):
What is the purpose of the multiple tsconfig files? Other than that and the things I pointed out, it all seems good.

ttoino:
> What is the purpose of the multiple tsconfig files? Other than that and the things I pointed out, it all seems good.

The typescript config is based on the one in create-vue, there's a file for the app itself and another one just for config files

Naapperas:
I'm of the opinion that props should be defined in a separate type, I believe it males the code easier to read. What do you think @ttoino ?

.github/workflows/ci.yml
ttoino: I don't think so, that step only checks if the current files follow the wanted style, it doesn't format them.

.github/workflows/ci.yml
ttoino: You're right, I'll look into it

ttoino:
> I'm of the opinion that props should be defined in a separate type, I believe it males the code easier to read. What do you think @ttoino ?

I think you're right, the current iteration of the typescript code leaves a lot to be desired and needs to be improved, especially on the prop typing.

Also, for some reason vue-tsc is not working right now as there are a lot of typing errors in vscode but the CI step is green 🤨

.github/workflows/ci.yml
Naapperas: I believe it previously used 'npm run format', this way it makes more sense

pedrosilva17 (dismissed):
Looks good to me. The tools chosen are adequate and the scripts will make our lives a lot easier when type checking and trying to maintain a consistent code format. Don't see anything wrong with the workflows either. Great job!

toni-santos:
LGTM, just missing the checks and should be good to merge! :+1:

pedrosilva17 (approved):
Same as before. LGTM 🐱‍👤

toni-santos (approved):

Naapperas (approved):

PR #34, opened 2023-08-21
Create Footer

### LINKED ISSUE #20: Create footer
We need a footer for the website, it should follow the designs provided by the design team. Furthermore it should also be responsive.

pedronunes19:
Closes #20 

The footer is created as a component following the mocks.  #25 should be merged for the icons.

Screenshot from 2023-08-21 18-19-23

### DIFF
--- a/resources/js/Components/Footer.vue
+++ b/resources/js/Components/Footer.vue
@@ -1,37 +1,30 @@
 <script setup>
+
+const contacts = [
+    ["io-logo-facebook", "bg-2023-red", "https://www.facebook.com/NIAEFEUP"],
+    ["io-logo-linkedin", "bg-2023-red-dark", "https://www.linkedin.com/company/nifeup/mycompany/"],
+    ["io-logo-instagram", "bg-2023-teal", "https://www.instagram.com/niaefeup/"],
+    ["io-mail", "bg-2023-orange", "mailto:ni@aefeup.pt"]
+];
 </script>
 
 <template>
     <footer class="bg-2023-bg py-4">
-        <section class="flex content-center items-center justify-center relative p-8 mx-[20%]">
-            <div class="flex gap-2 flex-row items-center w-full justify-evenly max-[900px]:hidden">
-                <a class="p-2.5 rounded-full flex w-fit bg-2023-red" target="_blank" href="https://www.facebook.com/NIAEFEUP">
-                    <v-icon name="io-logo-facebook" fill="white" scale="1.3"></v-icon>
-                </a>
-                <a class="p-2.5 rounded-full flex w-fit bg-2023-red-dark" target="_blank" href="https://www.linkedin.com/company/nifeup/mycompany/">
-                    <v-icon name="io-logo-linkedin" fill="white" scale="1.3"></v-icon>
-                </a>
-                <a class="p-2.5 rounded-full flex w-fit bg-2023-teal" target="_blank" href="https://www.instagram.com/niaefeup/">
-                    <v-icon name="io-logo-instagram" fill="white" scale="1.3"></v-icon>
-                </a>
-                <a class="p-2.5 rounded-full flex w-fit bg-2023-orange" target="_blank" href="mailto:ni@aefeup.pt">
-                    <v-icon name="io-mail" fill="white" scale="1.3"></v-icon>
-                </a>
+        <section class="flex content-center items-center justify-center relative p-8 mx-[25%]">
+            <div class="hidden gap-2 flex-row items-center w-full justify-evenly md:flex">
+                <template v-for="platform in contacts">
+                    <a class="p-2.5 rounded-full flex w-fit" :class="platform[1]" target="_blank" :href="platform[2]">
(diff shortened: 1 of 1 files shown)


Naapperas (dismissed):
LGTM

resources/js/Components/Footer.vue
```diff
@@ -0,0 +1,46 @@
...
+                </a>
+                <a class="p-2.5 rounded-full flex w-fit bg-2023-red-dark" target="_blank" href="https://www.linkedin.com/company/nifeup/mycompany/">
+                    <v-icon name="io-logo-linkedin" fill="white" scale="1.3"></v-icon>
+                </a>
+                <a class="p-2.5 rounded-full flex w-fit bg-2023-teal" target="_blank" href="https://www.instagram.com/niaefeup/">
+                    <v-icon name="io-logo-instagram" fill="white" scale="1.3"></v-icon>
+                </a>
+                <a class="p-2.5 rounded-full flex w-fit bg-2023-orange" target="_blank" href="mailto:ni@aefeup.pt">
+                    <v-icon name="io-mail" fill="white" scale="1.3"></v-icon>
+                </a>
```
ttoino: This repetition should be avoided with the use of components, or with a for loop

resources/js/Components/Footer.vue
```diff
@@ -0,0 +1,46 @@
...
+                    <a class="p-2.5 rounded-full flex w-fit bg-2023-red-dark" target="_blank" href="https://www.linkedin.com/company/nifeup/mycompany/">
+                        <v-icon name="io-logo-linkedin" fill="white" scale="1.3"></v-icon>
+                    </a>
+                    <a class="p-2.5 rounded-full flex w-fit bg-2023-teal" target="_blank" href="https://www.instagram.com/niaefeup/">
+                        <v-icon name="io-logo-instagram" fill="white" scale="1.3"></v-icon>
+                    </a>
+                    <a class="p-2.5 rounded-full flex w-fit bg-2023-orange" target="_blank" href="mailto:ni@aefeup.pt">
+                        <v-icon name="io-mail" fill="white" scale="1.3"></v-icon>
+                    </a>    
+                </div>
```
ttoino: Why are two divs necessary here?

resources/js/Components/Footer.vue
```diff
@@ -0,0 +1,46 @@
+<script setup>
+</script>
+
+<template>
+    <footer class="bg-2023-bg py-4">
+        <section class="flex content-center items-center justify-center relative p-8 mx-[20%]">
+            <div class="flex gap-2 flex-row items-center w-full justify-evenly max-[900px]:hidden">
```
ttoino: We should probably use tailwind's built-in breakpoints, not custom ones

ttoino (changes requested):

resources/js/Components/Footer.vue
pedronunes19: I don't think they're strictly necessary, it was just a away of separating the styles (mostly the `grid`). Follows pretty much the same structure that was being used for the maintenance page.

ttoino: While I don't really like it, it's an ok compromise as stuffing all classes in a single element would also be hard for readability

ttoino (approved):

resources/js/Components/Footer.vue
```diff
@@ -0,0 +1,39 @@
...
+    ["io-mail", "bg-2023-orange", "mailto:ni@aefeup.pt"]
+];
+</script>
+
+<template>
+    <footer class="bg-2023-bg py-4">
+        <section class="flex content-center items-center justify-center relative p-8 mx-[25%]">
+            <div class="hidden gap-2 flex-row items-center w-full justify-evenly md:flex">
+                <template v-for="platform in contacts">
+                    <a class="p-2.5 rounded-full flex w-fit" :class="platform[1]" target="_blank" :href="platform[2]">
```
Naapperas, line 16: The contacts could be objects, that would make the code a little cleaner. This is optional, so if you feel it does not make sense feel free to ignore this comment.

resources/js/Components/Footer.vue
```diff
@@ -0,0 +1,39 @@
...
+                    </a>
+                </template>
+            </div>
+            <div class="flex flex-col items-center gap-2 md:hidden">
+                <div class="grid grid-cols-2 gap-2 justify-center">
+                    <template v-for="platform in contacts">
+                        <a class="p-2.5 rounded-full flex w-fit" :class="platform[1]" target="_blank" :href="platform[2]">
+                            <v-icon :name="platform[0]" fill="white" scale="1.3"></v-icon>
+                        </a>
+                    </template>
```
Naapperas, line 27: Can't we have this logic separated? That way it is not duplicated (unless I saw things wrong but I double checked and believe they are the same for the mobile and desktop styles)

Naapperas (approved):
LGTM! @pedronunes19 I left some suggestions that can be ignored if you believe they don't make much sense.

PR #36, opened 2023-08-27
Create Navbar

pedrosilva17:
Closes #19 
This PR implements the website's navbar, both desktop and mobile versions. It is currently in its first iteration, since the code is not as squeaky clean as it could be, it's still missing icons (to be implemented in another PR soon) and the mobile version's design probably won't stay the same (When I started working on it, the mocks weren't complete). Moreover, some features are not yet implemented, for example, the action of each link in the navbar, since it requires the routes to exist and not all of them are being worked on yet. Nevertheless, take a look at the code and leave your suggestions below.

image
image
image

PR #37, opened 2023-08-29
Index Page

### LINKED ISSUE #33: Index page
Create the index page according to the mocks by the design team.

toni-santos:
Closes #33


toni-santos:
This is a WIP, most elements are not correctly aligned and are not yet mobile ready. Help and suggestions appreciated :)

toni-santos:
Desktop is mostly done, there is only a small issue that I've not been able to fix.
When scrolling through the speakers a small transition plays where a speaker's image increases in size whilst the current one becomes smaller, this makes the carousel height to change which makes the entire page's height shift. Any help regarding this issue is greatly appreciated!

Naapperas:
@toni-santos remember: se are already using Typescript, so we should.make.use of the type system. When defining props, teu to use a type as a genéric instead of an object.

resources/js/Components/Home/Map.vue
```diff
@@ -0,0 +1,43 @@
+<script setup lang="ts">
+import { LMap, LTileLayer, LMarker, LIcon } from "@vue-leaflet/vue-leaflet";
+import "leaflet/dist/leaflet.css";
+
+const zoom: number = 17;
+const center: Array<number> = [41.17835293313974, -8.595830311142494];
+const url: string = "https://tile.openstreetmap.de/{z}/{x}/{y}.png";
+const attribution: string =
+    '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors';
```
ttoino: Adding type hints here is redundant, the inferred types should be correct

resources/js/Components/Home/SpeakerSlide.vue
```diff
@@ -0,0 +1,31 @@
...
+}>();
+
+let active: boolean = props.slideID == props.currentSlide;
+
+watch(
+    () => props.currentSlide,
+    (newValue, oldValue) => {
+        active = newValue == props.slideID;
+    },
+);
```
ttoino: I think this can be replaced with a `reactive` or `computed` call, but I'm not sure

resources/js/Pages/Home.vue
```diff
@@ -0,0 +1,148 @@
...
+import Map from "../Components/Home/Map.vue";
+import SponsorBanner from "../Components/Home/SponsorBanner.vue";
+
+defineProps({
+    canLogin: Boolean,
+    canRegister: Boolean,
+    laravelVersion: String,
+    phpVersion: String,
+    modelValue: Object,
+});
```
ttoino: This should use typescript types, and `modelValue` should get a more descriptive type.

resources/js/app.ts
```diff
@@ -8,9 +8,12 @@ import { ZiggyVue } from "ziggy";
 import { OhVueIcon, addIcons } from "oh-vue-icons";
 
 import * as IoIcons from "oh-vue-icons/icons/io";
+import * as FaIcons from "oh-vue-icons/icons/fa";
 
 const Io = Object.values({ ...IoIcons });
+const Fa = Object.values({ ...FaIcons });
 addIcons(...Io);
+addIcons(...Fa);
```
ttoino, line 19: Do we need two icon families? Also, font-awesome is an aggregator of icons, so not all of their icons will look good together.

ttoino (changes requested):

CiscoPr:
I'm being quite nitpicky, but I believe in the footer, "niaefeup" should be capitalized as it represents the name of an organization, and it's predominantly presented this way on other platforms.
Captura de ecrã 2023-09-03 112341

resources/js/Components/Home/SpeakerSlide.vue
toni-santos: Nice catch, didn't know `computed` could be so useful!

resources/js/Pages/Home.vue
toni-santos: This was code left from the `Welcome.vue` file, which I used to make this one, that I forgot to remove.

resources/js/app.ts
toni-santos: font-awesome was added since some icons from io-icons do not behave correctly when setting their colors, as documented in [this issue](https://github.com/Renovamen/oh-vue-icons/issues/12) from a couple years ago.
Whilst it is possible to [add only specific icons](https://oh-vue-icons.js.org/docs), rather than the entire icon family, I believe it might be easier for other features where this issue arises to have the icon family available making it easier to use a more adequate icon.
Regardless, we must still be aware that io-icons is the main icon family, fa-icons acts only as a backup for these cases. Alternatively, we could ask the design team if switching the icon family is worth it.

pedronunes19:
> I'm being quite nitpicky, but I believe in the footer, "niaefeup" should be capitalized as it represents the name of an organization, and it's predominantly presented this way on other platforms. Captura de ecrã 2023-09-03 112341

That's my fault really (it's copied from the design team's mocks)

pedronunes19:
Screenshot from 2023-09-03 19-18-03

I believe you're aware of this @toni-santos but the alignment here seems a bit odd

resources/js/Pages/Home.vue
```diff
@@ -0,0 +1,140 @@
...
+                    Informática da Associação de Estudantes da Faculdade de
+                    Engenharia da Universidade do Porto (NIAEFEUP), foi criada
+                    com o intuito de permitir aos estudantes, independentemente
+                    do curso, desenvolver as suas capacidades nas diversas áreas
+                    da Informática, promovendo a sua interação com o mundo
+                    empresarial através de eventos sociais.
+                </p>
+                <p class="p-20 text-lg text-white max-lg:py-10">
+                    A edição de 2022 decorre entre os dias 31 de outubro a 2 de
+                    novembro. Os participantes têm a seu dispor palestras e
```
pedronunes19: We should probably change this, despite some other text (number of talks/workshops/etc.) needing a revision later

pedronunes19 (changes requested):

Naapperas:
@toni-santos I just thought of this: we should have the beginning and end dates not be hardcoded in the markup. They could be defined externally or added as attributes to an edition (in the last case, the website could just always use the latest edition available, regardless of whether it is active or not, food for thought)

Naapperas:
This is not scoped to this PR but I'll leave it here due to context; the dates part could/should still be handled here IMO.

Basically, if we had more attributes to an edition, we could define the site's theme as a JSON attribute (SQL standard has those now and PgSQL supports then already if I'm not mistaken) of the editon, as well as any other important data that might be needed.

toni-santos:
> Screenshot from 2023-09-03 19-18-03
> 
> I believe you're aware of this @toni-santos but the alignment here seems a bit odd

image

I'm not able to get a better screenshot, but I've changed the code so that the position of the titles, where applicable, are not computed by me but instead by CSS centering mechanisms which I then just add a margin in order to take into account the shadow when centering.

resources/js/Pages/Home.vue
toni-santos: This text must be revised once we are closer to the event, seeing as its information, alongside the ones from the "call to action" are more promotional/flavor rather than technical.

toni-santos:
> @toni-santos I just thought of this: we should have the beginning and end dates not be hardcoded in the markup. They could be defined externally or added as attributes to an edition (in the last case, the website could just always use the latest edition available, regardless of whether it is active or not, food for thought)

That's a great idea, in fact all pages should just be receiving information from the database and then displaying them through the mechanisms and interfaces that are respective to each edition. For this PR however, my idea was to keep it frontend only, seeing as the information being displayed here comes from sources from the database that are yet to be implemented in the backend, mainly, the speakers.

> This is not scoped to this PR but I'll leave it here due to context; the dates part could/should still be handled here IMO.
> 
> Basically, if we had more attributes to an edition, we could define the site's theme as a JSON attribute (SQL standard has those now and PgSQL supports then already if I'm not mistaken) of the editon, as well as any other important data that might be needed.

Agreed, and related to what I said regarding the first point :)

toni-santos:
I have fixed the issue I reported at the beginning where the carousel's changing height would make the rest of the page grow and shrink in size, however due to the fact that it made it impossible to achieve a clean transition I have disabled the animation/transition on the change of size of the image, furthermore it made the carousel look really weird when switching images. We could revisit this if we deem that the size transition is relevant and enhances the page, however due to its appearance I'd suggest against it.

Following this I believe there aren't anymore tweaks or fixes left. Please test the page out, mainly the speaker's carousel and its behavior when resizing and interacting with the page in responsive ways.

resources/js/Components/Home/SpeakersCarousel.vue
```diff
@@ -0,0 +1,76 @@
...
+        document.querySelector("ol.carousel__track");
+
+    if (carousel)
+        carousel.style.height = wrapperHeight ? wrapperHeight + "px" : "100%";
+};
+
+onMounted(async () => {
+    window.addEventListener("resize", wrapperHeight);
+
+    await new Promise((r) => setTimeout(r, 500)).then(wrapperHeight);
```
toni-santos, line 37: Forgot to mention, but from what I've been able to search and find regarding executing code on page load with Vue3, it seems that currently the `window.onload` [event is broken](https://github.com/vuejs/core/issues/5844), as `onMount` is being executed after the load/(not at all?) which means that we aren't able to wait for the full page load to get the correct height in `wrapperHeight`. As such this was the only way I found of running the function for the first time on page load, it is obviously less than ideal and any help is greatly appreciated.

pedronunes19 (dismissed):

resources/js/Components/Home/SpeakersCarousel.vue
Naapperas: I think you can set the timeout value to 0 and its still Will have the desired behavior: still better than waiting for some time

Naapperas: @ttoino thoughts?

toni-santos:
I may have not noticed or it may have been recently added, but there were some modals missing that provide some more information about the sponsors, this is now implemented with [this package](https://vue-final-modal.org/). It seems to be lightweight and easy to use, a good option whenever we need to use any modals!

resources/js/Components/Home/SpeakersCarousel.vue
toni-santos: I have attempted to use 0 as suggested but was unable to have the same behavior as with 500. I am befuddled by this problem...

Naapperas:
Inertia apparently has modal components

pedronunes19 (dismissed):

resources/js/Components/Home/Map.vue
```diff
@@ -0,0 +1,43 @@
...
+                <l-icon>
+                    <v-icon
+                        class="absolute -bottom-[9px] -right-[14px]"
+                        name="io-location-sharp"
+                        fill="#d94f04"
+                        scale="2"
+                    ></v-icon>
+                </l-icon>
+            </l-marker>
+        </l-map>
```
Naapperas: This is a matter of opinion but I would prefer if you use the components' name  in camel case instead. Makes it clearer that they are custom components.

Naapperas: Also, the maps looks a bit small in my opinion, I don't know if it would also look good if it were bigger.

resources/js/Components/Home/SpeakerSlide.vue
```diff
@@ -0,0 +1,26 @@
...
+</script>
+
+<template>
+    <div>
+        <img
+            src="https://picsum.photos/200"
+            class="w-[200px] rounded-full border-2 border-solid border-black object-cover"
+            :class="{ [`w-[400px]`]: active }"
+            alt=""
+        />
```
Naapperas: The transition between images should be smoother, right now the "next" picture just pops up instantly.

resources/js/Components/Home/SpeakerSlide.vue
```diff
@@ -0,0 +1,26 @@
...
+        <p v-if="active" class="py-4 text-lg text-2023-teal-dark">
+            John Doe {{ slideID }}
+        </p>
```
Naapperas: Not sure if the name should be disabled or simply smaller if the "speaker" slide is not active. Give it a try.

resources/js/Components/Footer.vue
```diff
@@ -62,7 +62,7 @@ const contacts = [
             </div>
         </section>
         <div class="container mx-auto text-center font-bold text-2023-teal">
-            <p>Copyright &copy; 2023 niaefeup, todos os direitos reservados</p>
+            <p>Copyright &copy; 2023 NIAEFEUP, todos os direitos reservados</p>
```
Naapperas: We could have a link in `NIAEFEUP` to our website.

CiscoPr (dismissed):

resources/js/Components/Home/SpeakerSlide.vue
```diff
@@ -0,0 +1,26 @@
...
+import { computed } from "vue";
+
+const props = defineProps<{
+    slideID: number;
+    currentSlide: number;
+}>();
+
+const active = computed(() => {
+    return props.currentSlide == props.slideID;
+});
```
Naapperas: This seems a bit off, slides should not need to be aware of the currently active slide. Perhaps they should just care if they are active or not and its up to their parent element to calculate that.

resources/js/Components/Home/SpeakersCarousel.vue
```diff
@@ -0,0 +1,76 @@
...
+    carousel.value?.next();
+};
+
+const prev = () => {
+    carousel.value?.prev();
+};
+
+const wrapperHeight = () => {
+    const wrapperHeight: number | undefined = document.querySelector(
+        "li.carousel__slide--active:nth-child(11) > div:nth-child(1)",
```
Naapperas: Not too keen on us depending on this: we have no guarantees that we will only have 10 speakers. This should come from a "through" relationship between speakers and editions, and there should be some fallback UI in case there are no speakers (as is the case now).

resources/js/Components/Home/SpeakersCarousel.vue
```diff
@@ -0,0 +1,76 @@
...
+    <Carousel
+        ref="carousel"
+        snap-align="center"
+        :autoplay="2000"
+        :items-to-show="2"
+        :wrap-around="true"
+        :breakpoints="breakpoints"
+        @slide-start="handleSlideStart"
+    >
+        <Slide v-for="slide in 10" :key="slide">
```
Naapperas, line 50: Same as above.

resources/js/Components/Home/SpeakersCarousel.vue
```diff
@@ -0,0 +1,76 @@
...
+            <SpeakerSlide :slideID="slide" :currentSlide="activeSpeaker" />
```
Naapperas: Regarding a comment made above, we should calculate here if this slide should be active or not.

resources/js/Components/Home/Sponsor.vue
```diff
@@ -0,0 +1,72 @@
+<script setup lang="ts">
+import { ref } from "vue";
+import { VueFinalModal, ModalsContainer } from "vue-final-modal";
+import "vue-final-modal/style.css";
+
+const getInitialValues = () => ({
+    modelValue: false,
+});
+
+const options = ref(getInitialValues());
```
Naapperas: Is there some reason for this code to be like this? Why not just set the ref directly?

resources/js/Components/Home/Sponsor.vue
```diff
@@ -0,0 +1,72 @@
...
+                    {{ props.description }}
+                    <br />
+                    Lorem ipsum dolor sit amet consectetur adipisicing elit.
+                    Natus adipisci voluptatem quod reprehenderit commodi nostrum
+                    ipsum eius eligendi qui vitae rerum dicta iure, officiis
+                    reiciendis totam veniam animi quidem laborum.
+                </p>
+            </div>
+        </VueFinalModal>
+        <ModalsContainer />
```
Naapperas: According to the [documentation](https://vue-final-modal.org/get-started/guide/setup#add-modalscontainer-to-vue-tree), you only need to add this component once, in a place that is common to every page that might make use of modals.

resources/js/Components/Home/Sponsor.vue
```diff
@@ -0,0 +1,72 @@
...
+    </div>
+</template>
+
+<style>
+.vfm__overlay {
+    background-color: rgba(248, 245, 231, 0.5);
+}
```
Naapperas: Should we mess with VFM directly like this?

resources/js/Components/Home/SponsorBanner.vue
Naapperas, line 1: I don't really like the way this was done here. I feel there's a lot of unnecessary repetition. We should simply have a component to render a group of sponsors, with these being passed in as props (this still needs to be thought about better). Other stuff such as color could also be passed in as props.

resources/js/Layouts/AppLayout.vue
```diff
@@ -27,9 +28,8 @@ defineProps<Props>();
             </header>
 
             <!-- Page Content -->
-            <main>
-                <slot />
-            </main>
```
Naapperas: Why did you remove the `main` element from here? Why did you move it elsewhere?

resources/js/Pages/Home.vue
```diff
@@ -0,0 +1,139 @@
+<script setup lang="ts">
+import AppLayout from "../Layouts/AppLayout.vue";
+import SpeakersCarousel from "../Components/Home/SpeakersCarousel.vue";
+import Map from "../Components/Home/Map.vue";
+import SponsorBanner from "../Components/Home/SponsorBanner.vue";
+</script>
+
+<template>
+    <AppLayout title="Home">
+        <main class="bg-2023-bg">
```
Naapperas: Same as above.

resources/js/Pages/Home.vue
```diff
@@ -0,0 +1,139 @@
...
+
+                <div class="relative">
+                    <img
+                        class="w-[450px] max-[900px]:w-[300px]"
+                        src="images/sinf logo.png"
+                        alt="Stylized SINF logo"
+                    />
+                    <span
+                        class="margin-0 absolute bottom-[-20px] right-0 text-xl font-bold text-2023-teal"
+                        >2023</span
```
Naapperas: This should come from the latest edition available, imo.

resources/js/Pages/Home.vue
```diff
@@ -0,0 +1,139 @@
...
+                    >
+                </div>
+                <p
+                    class="mr-2 border border-solid border-black p-2.5 px-8 text-lg font-bold text-2023-teal shadow-md shadow-2023-teal"
+                >
+                    semana_de_informática
+                </p>
+                <p class="margin-0 text-2xl font-bold text-2023-teal">
+                    25 a 31 de outubro
```
Naapperas: Same as above. We do not have that implemented yet, I think, but is something to keep in mind.

resources/js/Pages/Home.vue
```diff
@@ -0,0 +1,139 @@
...
+            <!-- SPONSORS -->
+            <section class="flex flex-col gap-10 px-20 py-24">
+                <p
+                    class="mr-[5px] flex w-min place-self-center border border-solid border-black bg-2023-teal-dark p-3 text-2xl font-bold text-white shadow shadow-2023-bg"
+                >
+                    Sponsors
+                </p>
+                <SponsorBanner tier="plat"></SponsorBanner>
+                <SponsorBanner tier="gold"></SponsorBanner>
+                <SponsorBanner tier="silver"></SponsorBanner>
```
Naapperas: This relates to the comment made above: the tiered sponsors should be injected into this View on page load: it is not the Banners responsibility to differentiate between them.

routes/web.php
```diff
@@ -16,7 +16,7 @@
 */
 
 Route::get('/', function () {
-    return Inertia::render('Welcome', [
+    return Inertia::render('Home', [
```
Naapperas: These props should be changed.

We should have information about:
- The edition
  - Its sponsors
  - Its speakers
  - Relevant dates

Naapperas (changes requested):
I saw a lot of `px` units and I'm kinda of scared: are they really necessary?

Other than that, I'm going to assume that you are waiting for #36 for the navbar: currently the "about us" section is half-cut and it could look better.

resources/js/Components/Home/SpeakersCarousel.vue
Naapperas: This is to say: this query looks kinda weird to me.

resources/js/Components/Home/Map.vue
toni-santos: I have kept these component's names according to their [documentation](https://github.com/vue-leaflet/vue-leaflet).

Increasing map size isn't problematic, seeing as the map is by default responsive and should not break anything else.

resources/js/Components/Home/SpeakerSlide.vue
toni-santos: I have discussed this issue [here](https://github.com/semanadeinformatica/website-2023/pull/37#issuecomment-1704390029).

resources/js/Components/Home/SpeakerSlide.vue
toni-santos: According to the Figma designs it should only appear when it is active. I could try this however, maybe we should talk with the design team regarding it.

resources/js/Components/Home/Map.vue
Naapperas: See if you can use different names

resources/js/Components/Home/SpeakerSlide.vue
toni-santos: This setup comes from the fact that there is no way to watch for the current active slide in the [carousel package we are using](https://ismail9k.github.io/vue3-carousel/getting-started.html). Therefore my solution was to make each `SpeakerSlide` know its ID and use `handleSlideStart` to update the current slide's value inside `SpeakerSlide`, changing the value of `active` that way.

resources/js/Components/Home/SpeakersCarousel.vue
toni-santos: I agree, I used the browser's CSS selector function for this, there is a better query.

resources/js/Components/Home/SpeakerSlide.vue
Naapperas: I've spoken with @madalenaye and she says we should add the names.

resources/js/Components/Home/Sponsor.vue
toni-santos: [This](https://vue-final-modal.org/use-cases/playground) was used as a basis. I'm not sure if it can be omitted.

resources/js/Components/Home/Sponsor.vue
toni-santos: VFM refers to Vue Final Modal, this is the [supposed usage](https://vue-final-modal.org/get-started/guide/setup#plain-css).

resources/js/Components/Home/Sponsor.vue
Naapperas: It can because, contrary to that example where they reuse the function, you only call it here. In fact, I can't even see the reason to use a ref here since you do not update its value, or at least nowhere that I could see.

resources/js/Components/Home/SpeakerSlide.vue
Naapperas: I understand that, but each slide just needs to know if it is active or not, not that it is slide X and currently slide Y is the active one. That calculation should come from the parent component.

toni-santos:
Branch has been rebased and is ready for review :) Please check it out!

Naapperas (dismissed):

routes/web.php
Naapperas: To be done in another PR.

resources/js/Pages/Home.vue
Naapperas: To be done in another PR.

resources/js/Pages/Home.vue
Naapperas: To be done in another PR.

resources/js/Components/Home/SpeakerSlide.vue
Naapperas: To be done in another PR.

Naapperas:
Don't forget to rebase this branch

ttoino (dismissed):
LGTM, I just have two very slight nitpicks:

At small screen sizes, maybe this title should be centered, and the grid reduced to one column.
image

Also at very small screen sizes, the speakers should either be smaller or the main speaker should have a higher z-index to appear on top.
image

Naapperas (approved):

ttoino (approved):

PR #38, opened 2023-08-30
Add admin features (CRUD)

### LINKED ISSUE #30: Implement admin functionalities
As the title suggests, we must start developing admin tools, namely CRUD operations for all models. Branch this issue into several ones as needed.

ttoino:
Closes #30.

The created pages can be found at /admin/<model>s. Example: /admin/users


Naapperas:
Is there anything missing in this PR @ttoino?

ttoino:
> Is there anything missing in this PR @ttoino?

I think it's done.

resources/js/Components/CRUD/Row.vue
```diff
@@ -0,0 +1,27 @@
+<script setup lang="ts">
+import type Model from "@/Types/Model";
+import Cell from "./Cell.vue";
+import { Link } from "@inertiajs/vue3";
+import route from "ziggy-js";
+
+defineProps<{
+    item: Model;
```
Naapperas, line 8: Why didn't you use a generic here like you did in the Table component?

Naapperas: Also, another nitpick: in other places you create an interface called Props but here you are defining the type inline, not really important but it's a matter of consistency.

Naapperas (dismissed):
LGTM. I believe that the pages should live under an `admin` folder and not a `CRUD` folder but that is not really that important right now.

Naapperas:
The tests are failing with the given message:
`Class "Illuminate\Support\Facades\App\Models\Student" not found`

Perhaps we messed up some paths, I believe this is not a problem with the logic itself of the website.

ttoino:
> The tests are failing with the given message: `Class "Illuminate\Support\Facades\App\Models\Student" not found`
> 
> Perhaps we messed up some paths, I believe this is not a problem with the logic itself of the website.

It was a problem with the factories, I added the usertype to the $with property, but the id was 0, causing an exception. It's now fixed.

resources/js/Components/CRUD/Row.vue
ttoino: I didn't use a generic because that particular element does not need any properties outside the `Model` interface and it doesn't expose that prop anywhere else so being a generic wouldn't be advantageous.

app/Http/Controllers/CRUDController.php
Naapperas, line 1: Maybe we could add a `getById` method to CRUD Controllers? I can see it being useful.

Naapperas: Besides that, you should write some documentation to explain what each thing does., for future reference.

app/Http/Controllers/CRUDController.php
```diff
@@ -0,0 +1,102 @@
...
+
+        return redirect()->action([static::class, 'index']);
+    }
+
+    public function destroy($id)
+    {
+        $this->model::find($id)->delete();
+
+        return redirect()->back();
+    }
```
Naapperas: Shouldn't we also have a `destroyed` method?

Naapperas (changes requested):
LGTM :rocket:

app/Http/Controllers/CRUDController.php
ttoino: > Maybe we could add a `getById` method to CRUD Controllers? I can see it being useful.

Sure, that'll be useful in the future if we need more in depth info for some models, like for validating quest requirements or forcing a user to renew their password.

> Besides that, you should write some documentation to explain what each thing does., for future reference.

Makes sense.

app/Http/Controllers/CRUDController.php
ttoino: I can't see how that would be useful, but I'll add it for completeness.

Naapperas: It might not be needed at all, its just a matter of completeness as you said

Naapperas (dismissed):

Naapperas (approved):

toni-santos (approved):

PR #44, opened 2023-09-05
Sticker component for sticker book

### LINKED ISSUE #40: Sticker book components
Create the components for the sticker book in the user's profile page.

This issue is related to #29. Designs are available on Figma.

toni-santos:
Closes #40.


resources/js/Components/Sticker.vue
```diff
@@ -0,0 +1,30 @@
...
+        </clipPath>
+    </svg>
+
+    <div class="group relative sticker-clip-path w-[220px] h-[220px] flex justify-center basis-0 overflow-hidden" :class="`bg-${color}` || {[`grayscale`] : !active}" >
```
ttoino: Even though `color` can be the empty string (and thus have a falsy value), the expression `bg-{color}` will always have a truthy value, making it the only value to ever be used.

One way of doing this could be like the following:

```suggestion
    <div class="group relative sticker-clip-path w-[220px] h-[220px] flex justify-center basis-0 overflow-hidden" :class="active ? `bg-${color}` || 'grayscale'" >
```

The syntax might not be right but you get the point. If I'm mistaken please correct me.

ttoino: As a bonus, we could look into the [clsx](https://www.npmjs.com/package/clsx) package on NPM to handle this kinds of things for us.

ttoino: Also, won't we run into problems for having the `class` attribute repeated? I genuinely don't know.

ttoino (changes requested):
By @Naapperas

resources/js/Components/Sticker.vue
toni-santos: I don't think that syntax would work, the goal is not to have an if statement where we check `active` and either apply the class `grayscale` or `bg-${color}`, but rather have the bind apply `bg-${color}` everytime and `grayscale` only when `!active` is true. I'm not sure if this is the behavior the design team intended but for a first draft should be good enough at showing if a sticker has been awarded or not.
Regarding the repeated `class` attribute, this is an intended use of Vue's [class/style bindings](https://vuejs.org/guide/essentials/class-and-style.html#binding-html-classes).

Naapperas: The link you provided mentions binding `class` to Arrays, take a look at those examples since I believe that is what you wanted to achieve.

resources/js/Components/Sticker.vue
```diff
@@ -0,0 +1,41 @@
...
+            />
+        </clipPath>
+    </svg>
```
ttoino, line 17: This svg should only be rendered once, this can be achieved with the Head component as is highlighted in the Inertia docs: https://inertiajs.com/title-and-meta#multiple-head-instances

toni-santos: This did not seem to work, not only was the svg still being added multiple times, only this time in the head.
image psd

ttoino: Did you add the head-key attribute?

toni-santos: In the screenshot I didn't but I tried and it did only create a single svg, however I wasn't able to select the clipPath. A simple fix, and given that this component will only be used on a single page, we can include the svg in its parent container and all stickers will be able to use that svg. What do you think about this?

Naapperas (dismissed):

Naapperas (approved):

ttoino (approved):

PR #49, opened 2023-09-12
Added maintenance mode script

Naapperas:
This PR adds some QOL scripts that can aid in the development of the website.

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -24,7 +24,7 @@ class Handler extends ExceptionHandler
             return Inertia::render('Error', ['status' => $response->status()])
                 ->toResponse($request)
                 ->setStatusCode($response->status());
-        } elseif ($response->status() === 503 && app()->environment(['maintenance'])) {
+        } elseif ($response->status() === 503) {
             return Inertia::render('Maintenance')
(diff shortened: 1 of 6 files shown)


app/Exceptions/Handler.php
```diff
@@ -24,7 +24,7 @@ public function render($request, Throwable $e)
             return Inertia::render('Error', ['status' => $response->status()])
                 ->toResponse($request)
                 ->setStatusCode($response->status());
-        } elseif ($response->status() === 503 && app()->environment(['maintenance'])) {
+        } elseif ($response->status() === 503) {
```
toni-santos, line 27: This is not necessarily a problem, as we will most likely only use it under maintenance, but 503 is not necessarily only for maintenance.

Naapperas: I made this change since it disrupted a bit the "run the maintenance script and the site is in maintenance mode" flow. If you still feel that this should not be the case I'll rollback this change tho.

Naapperas:
Had to change the migration because it would raise errors when rolling back.

toni-santos (approved):

pedronunes19 (approved):

PR #52, opened 2023-09-13
Added data sharing agreement checkbox to registration form

Naapperas:


pedrosilva17 (approved):

resources/js/Components/PrimaryButton.vue
```diff
@@ -2,7 +2,7 @@
 import type { ButtonHTMLAttributes } from "vue";
 
 interface Props {
-    type: ButtonHTMLAttributes["type"];
+    type?: ButtonHTMLAttributes["type"];
```
ttoino, line 5: Does this change anything? I thought `withDefaults` already made it optional.

ttoino (approved):

resources/js/Components/PrimaryButton.vue
Naapperas: Just some console warnings regarding required props: even though we have a default value, the type itself does not allow for the prop to be undefined, hence the warnings. The Vue documentation for this feature has the attributes as optional in the original type. I just changed it here because it was directly related to the form I was editing and, honestly, was giving me an ick.

Naapperas: From what I understood, if the Prop type has an attribute that can be `undefined`, `withDefaults` basically "switches" the `undefined` to the type of the default value.

PR #67, opened 2023-09-15
Added missing sponsor relationships

Naapperas:


pedrosilva17 (dismissed):

CiscoPr (dismissed):

app/Models/Edition.php
```diff
@@ -39,4 +40,9 @@ public function quests(): HasMany
     {
         return $this->hasMany(Quest::class);
     }
+
+    public function sponsors(): BelongsToMany
+    {
+        return $this->belongsToMany(Sponsor::class);
+    }
```
ttoino: Shouldn't this be a hasMany relationship?

ttoino: Looking at the docs, belongsToMany is used in many to many relations, and, from what I understand, sponsors can only belong to one edition, right?

app/Models/Company.php
```diff
@@ -31,4 +32,9 @@ public function user(): BelongsTo
     {
         return $this->belongsTo(User::class);
     }
+
+    public function sponsors(): BelongsToMany
+    {
+        return $this->belongsToMany(Sponsor::class);
+    }
```
ttoino: Same as the other comment.

app/Models/Edition.php
Naapperas: `Sponsor` is the association class representing the "Many to Many" association between `Company` and `Edition`. I gave a quick look at the documentation but honestly couldn't quite figure out what should be used here: perhaps you're right.

Naapperas: But to answer your question, yes, sponsors only belong to a single edition and represent a single company

ttoino: Laravel does not allow association classes like that, you're defining the relationship between companies and sponsors, so the correct one would be hasMany I think

Naapperas: On it.

ttoino (approved):

diogofonte (approved):
Looks good!

pedrosilva17 (approved):
I was convinced `belongsToMany()` represented a one to many relationship. Thank you @ttoino 👍

PR #69, opened 2023-09-15
Inject database models in home page

ttoino:
Closes #54

PR #71, opened 2023-09-16
Improve admin tools

### LINKED ISSUE #58: Improve admin tools
The current admin controllers are very generic and allow creating, updating and deleting models, but reading is limited to a list of models. We should add pages for individual models where appropriate. This page could show some more information, like the number of users enrolled into an edition.

We should add additional functionality where required, like requesting a user password change or maybe even sending a custom email to a particular user.
 
Editing a users picture is also impossible right now.

ttoino:
Related to #58.

### DIFF
--- a/app/Http/Controllers/UserCRUDController.php
+++ b/app/Http/Controllers/UserCRUDController.php
@@ -87,9 +87,11 @@ class UserCRUDController extends CRUDController
         }
 
         if ($new['type'] !== 'admin' && isset($new['social_media'])) {
-            $socialMedia = $old->social_media;
+            $socialMedia = $old->usertype->social_media;
             if ($socialMedia === null) {
                 $socialMedia = SocialMedia::create($new['social_media']);
+                $old->usertype->socialMedia()->associate($socialMedia);
+                $old->usertype->save();
             } else {
                 $socialMedia->update($new['social_media']);
             }
@@ -102,7 +104,6 @@ class UserCRUDController extends CRUDController
         return [
             'name' => $new['name'],
             'email' => $new['email'],
-            'social_media_id' => $socialMedia?->id ?? null,
         ];
     }
 }
--- a/app/Http/Controllers/UserCRUDController.php
+++ b/app/Http/Controllers/UserCRUDController.php
@@ -26,6 +26,7 @@ class UserCRUDController extends CRUDController
         'social_media.linkedin' => 'sometimes|nullable|string',
         'social_media.twitter' => 'sometimes|nullable|string',
         'social_media.website' => 'sometimes|nullable|string|url',
+        'photo' => 'nullable|mimes:jpg,jpeg,png|max:1024',
     ];
 
     protected array $search = ['name', 'email'];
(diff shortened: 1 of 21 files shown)


Naapperas (dismissed):

Naapperas (dismissed):

toni-santos (dismissed):

app/Http/Controllers/UserCRUDController.php
```diff
@@ -56,31 +66,39 @@ protected function created(array $new): ?array
...
 
-        if ($old['usertype_type'] !== $type) {
+        if ($old->usertype_type !== $type) {
             // This should never happen
             throw new \Exception('Cannot change user type');
         }
 
         if ($new['type'] !== 'admin' && isset($new['social_media'])) {
-            $socialMedia = SocialMedia::find($old['social_media_id']);
+            $socialMedia = $old->social_media;
```
pedrosilva17: This is always equal to null, since the $old array doesn't have a social_media_id associated with it, I believe we need to query the appropriate model (Student or Company) using the usertype_id. This is the reason why social media isn't being updated: as the result is always null, a new row in the SocialMedia table is constantly being created, but the social_media_id isn't updated with it (unsure why this happens, but it shouldn't need to be updated regardless, as one user should only have one entry in the SocialMedia table). As such, the social_media_id needs to be retrieved from either table, depending on the user type.

pedrosilva17 (commented):
Some things I've noticed:
- I'm not sure if it's only happening with me, but after editing a user's picture, it no longer appears when I'm in the edit page for that user.
- Creating a new Company causes the following error: 
image
- Updating social media currently has no effect. More on it in the comments
- `isset()` returns `true` when evaluating an array of keys with null values, which in this case means it will always return true, regardless of whether I type anything in the social media inputs or not (for example), and it will also create those types of arrays in the database. Not sure if this was intended or not, my idea was that an entry in the SocialMedia table would only be created/updated if anything was typed in the social media inputs. However, if this behavior is fixed, maybe we would need to remove that check when updating, as I might want to delete all of my social media from the website (and that will be impossible if arrays of keys with null values are properly filtered).

ttoino:
> Some things I've noticed:
> 
> * I'm not sure if it's only happening with me, but after editing a user's picture, it no longer appears when I'm in the edit page for that user.

This is caused by two things, firstly, you have to run the `storage:link` artisan command to create the relevant folders where pfps are stored, secondly, you need to change the `APP_URL` env variable to include the port (this is already done in #69).

> * Creating a new Company causes the following error:
>   image

This works for me. You probably need to migrate the db.

> * Updating social media currently has no effect. More on it in the comments
> * `isset()` returns `true` when evaluating an array of keys with null values, which in this case means it will always return true, regardless of whether I type anything in the social media inputs or not (for example), and it will also create those types of arrays in the database. Not sure if this was intended or not, my idea was that an entry in the SocialMedia table would only be created/updated if anything was typed in the social media inputs. However, if this behavior is fixed, maybe we would need to remove that check when updating, as I might want to delete all of my social media from the website (and that will be impossible if arrays of keys with null values are properly filtered).

~~I'll look into these issues.~~ This should now be fixed.

pedrosilva17 (approved):

toni-santos (approved):

PR #78, opened 2023-09-18
Landing page for admins

### LINKED ISSUE #73: Admin landing page
We need a simple landing page for admins so that they do not need to navigate the site's admin features via URLs.

toni-santos:
Closes #73

### DIFF
--- a/resources/js/Pages/Admin.vue
+++ b/resources/js/Pages/Admin.vue
@@ -1,5 +1,4 @@
 <script setup lang="ts">
-import PrimaryButton from "@/Components/PrimaryButton.vue";
 import AdminLayout from "@/Layouts/AdminLayout.vue";
 import { computed } from "vue";
 import route from "ziggy-js";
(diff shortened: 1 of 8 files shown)


resources/js/Pages/Admin.vue
```diff
@@ -0,0 +1,62 @@
...
+    },
+    "admin.departments.index": {
+        label: "Departments",
+    },
+    "admin.stands.index": {
+        label: "Stands",
+    },
+    "admin.eventDays.index": {
+        label: "Event Days",
+    },
```
ttoino: These should probably be in portuguese

routes/web.php
```diff
@@ -61,6 +61,10 @@ function () {
                     '/stands' => StandCRUDController::class,
                     '/eventDays' => EventDayCRUDController::class,
                 ]);
+
+                Route::name('index')->get('/index', function () {
```
ttoino: I'd prefer `/`

ttoino (changes requested):

pedrosilva17:
Agree with João's observations. Other than that, useful landing page 👍

pedrosilva17:
I just remembered, maybe we should have a section for managing enrollments? We could change the amount of points of a participant in case a problem arises, just in case

ttoino (approved):

Naapperas (approved):

PR #48, opened 2023-09-10
Student Profile Page

pedronunes19:
Closes #29

PR #88, opened 2023-09-19
Company Profile Page 

### LINKED ISSUE #86: Company Profile Page
We must create a separate page for the companies, with the following:

- basic info of the company (just like the normal student page)
- list of the students which interacted with them
-  the table should have the students' **name**, **mail**, **LinkedIn** and **CV**
- if the company is **SILVER** tier then, instead of a table with all the info of the students, it can **only** have the user's name and mail

### LINKED ISSUE #196: Fix team page social links
Currently they use the same tab when redirecting to the given social link page. They should open a new tab. This basically means adding a `target="_blank"` to the links?

CiscoPr:
Closes #86
Closes #196


ttoino:
Ideally this would use the same route as in #48.

CiscoPr:
> Ideally this would use the same route as in #48.

Yeah, i will refactor that

Naapperas:
Don't forget to format the code

resources/js/Pages/Profile/Show.vue
```diff
@@ -23,41 +24,23 @@ defineProps<Props>();
...
                 </div>
-            </template>
+
+                <CvArea
+                    v-if="
+                        $page.props.auth.user.usertype_type !==
+                        'App\\Models\\Company'
+                    "
+                    :item="user ?? $page.props.auth.user"
+                />
```
pedronunes19: this should only appear when either user.usertype_type is Participant (if it exists) or the auth user is Participant

pedronunes19 (changes requested):

resources/js/Pages/Profile/Show.vue
CiscoPr: yeah, i forgot some cases for this condition, ty

Naapperas:
What is missing? The student visit history?

CiscoPr:
> What is missing? The student visit history?

are you talking about the table?

Naapperas:
> > What is missing? The student visit history?
> 
> are you talking about the table?

I'm talking about what features or fixes are missing for this PR to be merged.

CiscoPr:
> > > What is missing? The student visit history?
> > 
> > 
> > are you talking about the table?
> 
> I'm talking about what features or fixes are missing for this PR to be merged.

what is missing is to get the participants correctly... the relationship as it is for now always returns an undefined array

Naapperas:
> > > > What is missing? The student visit history?
> > > 
> > > 
> > > are you talking about the table?
> > 
> > 
> > I'm talking about what features or fixes are missing for this PR to be merged.
> 
> what is missing is to get the participants correctly... the relationship as it is for now always returns an undefined array

Due to the way quests are linked to stands, we may need more complex logic than just having a deep query.

At what point does the query fail? That's the point where we need to act upon.

resources/js/Pages/Profile/Show.vue
```diff
@@ -9,12 +9,15 @@ import TwoFactorAuthenticationForm from "@/Pages/Profile/Partials/TwoFactorAuthe
...
+import type { User } from "@/Types/User";
+import EnrolledParticipants from "@/Components/Profile/EnrolledParticipants.vue";
 import type Participant from "@/Types/Participant";
 
 interface Props {
     confirmsTwoFactorAuthentication: boolean;
     sessions: Session[];
-    participant: Participant;
+    user: User;
+    participants: Participant[];
```
pedronunes19: I don't think this prop is being passed when we use the default route from jetstream, that's likely why it's always undefined (one of the reasons at least)

CiscoPr: You made a good point. However, i already had noted that... i just forgot to push :,))
And yet again, this doesn't solve the problem

app/Http/Controllers/UserController.php
```diff
@@ -15,41 +19,126 @@ public function show(Request $request)
...
+        $user = $requestUser;
+
+        /** @var User|null */
+        $paramUser = User::find($request->route('user'));
+
+        if ($paramUser !== null) {
+            if ($requestUser->cannot('viewProfileOf', $paramUser)) {
+                // If the currently logged in user is not an admin, they can only view their own profile (or other users if they are a company and the user is a participant of that company)
+
+                return redirect()->route('profile.show')->dangerBanner('Não tens acesso a esta página.');
```
toni-santos, line 89: Why do this instead of just showing 403?

app/Providers/AuthServiceProvider.php
```diff
@@ -68,5 +68,35 @@ public function boot(): void
...
+                $user->isCompany() &&
+                $profile_user->isParticipant() &&
+                $user->usertype->participants()->exists($profile_user)
+            )
+        ));
+
+        Gate::define('viewCVOf', fn (User $user, User $cv_user, Edition $edition) => (
+            ( // admins have access to all CVs that do not belong to admins (since they do not have those)
+                $user->isAdmin() &&
+                ! $cv_user->isAdmin()
```
toni-santos: What do you mean by "that do not belong to admins"?

app/Http/Controllers/UserController.php
Naapperas: :boom: error page :boom:

app/Providers/AuthServiceProvider.php
Naapperas: Perhaps my wording was off here: this was just a way of securing this validation agains an admin trying to view another admin's profile.

CiscoPr:
we should be careful, because, when clicking on this on a student profile as a company, it will return this error:
image
image

CiscoPr:
on another note, we should also increase the width of this CV button in the same situation (company visiting enrolled student page)
image

CiscoPr:
Finally, i think we should also decrease the size of the "no students enrolled yet" message. Imo, it's too big

pedronunes19:
> we should be careful, because, when clicking on this on a student profile as a company, it will return this error: image image

Why can a company even see this?

Naapperas:
> > we should be careful, because, when clicking on this on a student profile as a company, it will return this error: image image
> 
> Why can a company even see this?

I believe companies should at least be able to see what events the user has joined that may be related to them

app/Http/Controllers/UserController.php
Naapperas: I'm marking this as resolved but if you feel that this should be changed go ahead.

pedronunes19 (dismissed):

Naapperas (dismissed):

toni-santos (dismissed):

pedronunes19 (approved):

ttoino (approved):

PR #79, opened 2023-09-18
Added competition info

Naapperas:
Closes #45

PR #91, opened 2023-09-21
Add quest slots

### LINKED ISSUE #41: Sticker book backend logic
The sticker book is a way to celebrate and incentivise the users to explore and participate in the event, however the backend logic necessary to award the stickers is missing. This is tied to quests and their completion.

Naapperas:
Closes #41. This PR basically changes the way we count points on an enrollment.

The idea is that there are slots that can be "filled" by 1-N quests, when a user completes a quest, this slot can be marked completed or not: if it is completed it is treated as "unlocked" and made available to the enrollment.


Naapperas:
This should only be merged after #79 is merged.

toni-santos (dismissed):
Looks good, we could maybe add a secret "ingredient" [Count Dracula voice]!

CiscoPr (approved):

toni-santos (approved):

PR #105, opened 2023-09-22
Add event type and make speaker a user type

### LINKED ISSUE #100: Add event type and make speakers a usertype
Events should have an enum specifying their type, like workshop, talk, group dynamic, etc.

Speakers should be a user type, like companies and participants. This simplifies some logic in events, as some events also have companies.

ttoino:
Closes #100


app/Http/Controllers/ProgramController.php
```diff
@@ -19,8 +19,7 @@ public function show(Request $request)
         }
 
         // We can do this here since we already fetched the edition from the DB, so we would not gain any performance from "short-circuiting" the validation
```
Naapperas: Please remove this comment as it no longer makes sense.

app/Http/Controllers/UserCRUDController.php
```diff
@@ -19,7 +20,9 @@ class UserCRUDController extends CRUDController
         'name' => 'required|string|max:255',
         'email' => 'required|string|email|max:255|unique:users,email',
         'type' => 'required|in:participant,company,admin',
+        'title' => 'sometimes|nullable|string',
         'description' => 'sometimes|nullable|string',
+        'organization' => 'sometimes|nullable|string',
```
Naapperas, line 25: Perhaps we could use the conditional validation rules provided by native Laravel.

Naapperas (changes requested):

app/Http/Controllers/UserCRUDController.php
ttoino: As these options are optional, they wouldn't really help here I think

Naapperas (approved):
:+1:

toni-santos (approved):

PR #113, opened 2023-09-22
Started reworking program page

### LINKED ISSUE #99: Show event day theme only for talks

Naapperas:
Closes #99


Naapperas:
This PR is waiting for #105

app/Traits/HasCV.php
```diff
@@ -19,6 +19,8 @@ public function updateCV(UploadedFile $cv, $storagePath = 'cvs')
         tap($this->cv_path, function ($previous) use ($cv, $storagePath) {
             if ($previous) {
                 Storage::disk($this->CVDisk())->delete($previous);
+            } else {
+                // Give points for CV submission
```
pedronunes19, line 23: So we're still not allowing users to delete their CV, right?

resources/js/Components/Navbar.vue
```diff
@@ -91,7 +95,10 @@ const isAdmin = computed(() => {
                         <DropdownLink
                             :href="route('competition.show', { competition })"
                         >
-                            {{ competition.name }}
+                            {{
+                                competition.name ??
+                                `Competition ${competition.id}`
+                            }}
```
pedronunes19: why are we allowing null competition names?

pedronunes19:
How are we dealing with other events like the panel, for example?

Naapperas:
> How are we dealing with other events like the panel, for example?

That could be another event type. CC @ttoino

resources/js/Components/Navbar.vue
Naapperas: We are not, I don't know why I put that there but it's fixed now.

app/Traits/HasCV.php
Naapperas: For now at least.

ttoino:
> > How are we dealing with other events like the panel, for example?
> 
> That could be another event type. CC @ttoino

It could be another type or it could just be a talk, we support multiple speakers already.

toni-santos (approved):

PR #121, opened 2023-09-23
Add Dockerfile for prod

ttoino:


Dockerfile
Naapperas, line 1: This Dockerfile needs to be heavily refactored but for now it fits our needs so no worries there.

Naapperas (approved):

toni-santos (approved):

PR #127, opened 2023-09-26
Add way for users to enroll

### LINKED ISSUE #125: Add enrollment backend
Add a way for participants to enroll in an edition

ttoino:
Closes #125.

### DIFF
--- a/resources/js/Components/Home/EnrollSection.vue
+++ b/resources/js/Components/Home/EnrollSection.vue
@@ -31,7 +31,11 @@ import route from "ziggy-js";
                 shadow="teal"
                 text-size="sm:text-4xl"
                 padding="sm:px-8"
-                @click="router.put(route('enroll'))"
+                @click="
+                    $page.props.auth.user
+                        ? router.put(route('enroll'))
+                        : router.get(route('register'))
+                "
                 >Inscrever-me</PrimaryButton
             >
         </div>
--- a/resources/js/Components/Home/EnrollSection.vue
+++ b/resources/js/Components/Home/EnrollSection.vue
@@ -27,7 +27,7 @@ import PrimaryButton from "@/Components/PrimaryButton.vue";
             <PrimaryButton
                 color="orange"
                 shadow="teal"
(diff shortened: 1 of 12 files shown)


Naapperas (dismissed):

resources/js/Pages/Home.vue
```diff
@@ -179,7 +180,7 @@ const formattedDate = (
             ></SponsorBanner>
         </section>
         <!-- CALL TO ACTION -->
-        <EnrollSection />
+        <EnrollSection v-if="canEnroll" />
```
pedrosilva17, line 183: Do we want the call to action hidden to unregistered users?

toni-santos: I think it should be hidden, registered users could be confused by the button.

toni-santos (dismissed):
LGTM :+1:

resources/js/Pages/Home.vue
pedrosilva17: In my opinion, the condition should verify if the user is not logged in or, when logged in, if the user can enroll. In case the user clicks the button while not logged in, the action should redirect to the login page.

toni-santos: I misunderstood your comment you're right. Users that are not logged in aren't being shown the button, the condition should check for log in and registration on the event in order to hide it. Redirecting the user to log in should also be done in case they aren't.

resources/js/Components/Home/EnrollSection.vue
```diff
@@ -27,8 +29,9 @@ import PrimaryButton from "@/Components/PrimaryButton.vue";
             <PrimaryButton
                 color="orange"
                 shadow="teal"
-                textSize="sm:text-4xl"
+                text-size="sm:text-4xl"
                 padding="sm:px-8"
+                @click="router.put(route('enroll'))"
```
toni-santos: As discussed previously.

pedrosilva17 (approved):
Good job! 👍

toni-santos (approved):

PR #129, opened 2023-09-26
Fix navbar for desktop and mobile

toni-santos:
Redesign the navbar and fix smaller related issues.

### DIFF
--- a/resources/js/Components/HamburgerMenu.vue
+++ b/resources/js/Components/HamburgerMenu.vue
@@ -23,11 +23,7 @@ const props = defineProps<{
 }>();
 
 watch(open, () => {
-    if (open.value) {
-        document.body.classList.add("overflow-hidden");
-    } else {
-        document.body.classList.remove("overflow-hidden");
-    }
+    document.body.classList.toggle("overflow-hidden");
 });
 </script>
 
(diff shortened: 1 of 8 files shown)


pedronunes19 (dismissed):
./format.sh

pedronunes19:
Also the previous editions dropdown doesn't seem very intuitive to me compared to previous iterations but I guess it's consistent with other dropdowns

toni-santos:
This should be ready, we should just check everything with the design team before merging.

resources/js/Components/HamburgerMenu.vue
```diff
@@ -21,6 +21,14 @@ const props = defineProps<{
         editions: number[];
     };
 }>();
+
+watch(open, () => {
+    if (open.value) {
+        document.body.classList.add("overflow-hidden");
+    } else {
+        document.body.classList.remove("overflow-hidden");
+    }
```
Naapperas:
```suggestion
    document.body.classList.toggle("overflow-hidden");
```

Isn't this a cleaner alternative?

resources/js/Components/HamburgerMenu.vue
```diff
@@ -92,14 +100,18 @@ const props = defineProps<{
...
+                style="grid-template-columns: repeat(auto-fill, 100px)"
+            >
                 <template
                     v-for="edition in props.options.editions"
                     :key="`navbar-edition-link-${edition}`"
                 >
                     <ResponsiveNavLink
                         :href="route('home')"
                         :active="edition === new Date().getFullYear()"
+                        class="!w-fit"
```
Naapperas: Why the `!`?

Naapperas (changes requested):
Overall looks good, however the logo seems a little bit too detailed to just be a "side bar element".

Other than that, maybe we should add smooth-scrolling to the website? (most noticeable when accessing home page links)

Naapperas (dismissed):

Naapperas (approved):

PR #132, opened 2023-09-26
Shop

### LINKED ISSUE #131: Shop page
We must have a shop where users are able to trade their points for goodies.

toni-santos:
Closes #131

### DIFF
--- a/app/Http/Controllers/ShopController.php
+++ b/app/Http/Controllers/ShopController.php
@@ -2,7 +2,6 @@
 
 namespace App\Http\Controllers;
 
-use App\Models\Event;
 use App\Models\Product;
 use Illuminate\Http\Request;
 use Inertia\Inertia;
--- a/app/Http/Controllers/ShopController.php
+++ b/app/Http/Controllers/ShopController.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use App\Models\Event;
+use App\Models\Product;
+use Illuminate\Http\Request;
+use Inertia\Inertia;
+
+class ShopController extends Controller
+{
+    public function show(Request $request)
+    {
+        $edition = $request->input('edition');
+
+        if ($edition === null) {
+            return response('No edition found', 500);
+        }
+
+        $products = Product::where('edition_id', $edition->id)->get();
+
+        return Inertia::render('Shop', [
+            'products' => $products,
+        ]);
+    }
+}
--- a/app/Http/Controllers/ShopController.php
+++ b/app/Http/Controllers/ShopController.php
@@ -2,7 +2,6 @@
 
 namespace App\Http\Controllers;
 
-use App\Models\Product;
 use Illuminate\Http\Request;
 use Inertia\Inertia;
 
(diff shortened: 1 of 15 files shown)


resources/js/Components/Shop/ShopItem.vue
```diff
@@ -0,0 +1,25 @@
+<script setup lang="ts">
+import { defineProps } from 'vue';
+
+interface Props {
+    name: string;
+    price: number;
+    description: string;
+    image: string;
+}
```
Naapperas: We already have the Product type: you should use that, even when mocking.

resources/js/Pages/Shop.vue
```diff
@@ -0,0 +1,44 @@
...
+    name: 'Item 1',
+    price: 100,
+    description: 'Lorem ipsum dolor sit amet consectetur adipisicing elit. Quisquam, voluptatum.',
+    image: 'https://picsum.photos/200/200'
+},{
+    name: 'Item 1',
+    price: 100,
+    description: 'Lorem ipsum dolor sit amet consectetur adipisicing elit. Quisquam, voluptatum.',
+    image: 'https://picsum.photos/200/200'
+},]
```
Naapperas: Same as above.

Naapperas (changes requested):

app/Http/Controllers/ShopController.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    public function show(Request $request)
+    {
+        $edition = $request->input('edition');
+
+        if ($edition === null) {
+            return response('No edition found', 500);
+        }
+
+        $products = Product::where('edition_id', $edition->id)->get();
```
Naapperas:
```suggestion
        $products = $edition->products;
```

This is more idiomatic, in my opinion.

Naapperas: You could even inline this in the return value

resources/js/Components/Shop/ShopItem.vue
Naapperas, line 1: Are you going to implement buying items? Nonetheless, we should already have a button for that. If you have other ideas please share them.

Naapperas (changes requested):
There is no way for a user to go to the shop unless they know the exact Url, maybe a link in the Navbar?

Naapperas (dismissed):

pedronunes19:
> There is no way for a user to go to the shop unless they know the exact Url, maybe a link in the Navbar?

Was this fixed?

Naapperas:
> > There is no way for a user to go to the shop unless they know the exact Url, maybe a link in the Navbar?
> 
> Was this fixed?

I believe it was, but if it was not request for that to be changed.

toni-santos:
> > > There is no way for a user to go to the shop unless they know the exact Url, maybe a link in the Navbar?
> > 
> > 
> > Was this fixed?
> 
> I believe it was, but if it was not request for that to be changed.

This hasn't been addressed has #129 introduces an altered navbar and hasn't been merged.

pedronunes19 (approved):

PR #136, opened 2023-09-26
Edit profile page

### LINKED ISSUE #108: Add page for users to edit their info
Should allow users to change their name/email and social media.

pedrosilva17:
Closes #108 
Currently similar to the default edit profile page from Laravel, now in a different route. The sentences in here need to be added to #75 (I can do that after this is merged, if you want @Naapperas)

image

image

image


resources/js/Pages/Profile/Edit.vue
```diff
@@ -0,0 +1,73 @@
...
+        <template #header>
+            <h2
+                class="text-xl font-semibold leading-tight text-gray-800 dark:text-gray-200"
+            >
+                Profile
+            </h2>
+        </template>
+
+        <div>
+            <div class="mx-auto max-w-7xl py-10 sm:px-6 lg:px-8">
```
Naapperas: I believe `7xl` is a lot in this case: `2xl` or `3xl` should be enough.

Naapperas (changes requested):
Please be careful with the amount of `div`s you use. Ask @ttoino if you have any doubts regarding that matter.

Naapperas (dismissed):

Naapperas:
@pedrosilva17 one thing I noticed: perhaps we could make the font of this page's text bold, or at least make it stand out more.

Naapperas (approved):

toni-santos (approved):

Issue #116, opened 2023-09-23
Profile page "tickets" and "stickers" implementation 


toni-santos:
Now that we have the profile page ready on the main branch, we can implement the missing pieces, the "sticker book" and "tickets".

PR #139, opened 2023-09-28
Added initial backend code for event registration

### LINKED ISSUE #43: Activity sign up backend logic
Users should be able to sign up for an activity. It's important to note that activities may have limited availability and are time sensitive, that is, an activity can only be signed up for some amount of time before it starts and it may be already fully booked.

Naapperas:
Closes #43

### DIFF
--- a/app/Http/Controllers/EventController.php
+++ b/app/Http/Controllers/EventController.php
@@ -4,7 +4,6 @@ namespace App\Http\Controllers;
 
 use App\Models\Event;
 use Illuminate\Http\Request;
-use Illuminate\Support\Facades\Gate;
 use Inertia\Inertia;
 
 class EventController extends Controller
@@ -12,7 +11,7 @@ class EventController extends Controller
     public function show(Event $event)
     {
         return Inertia::render('Event', [
-            'event' => $event->load(['users', 'event_day']),
+            'event' => $event->load(['users', 'event_day', 'enrollments']),
         ]);
     }
 
@@ -21,7 +20,12 @@ class EventController extends Controller
      */
     public function join(Request $request, Event $event)
     {
-        Gate::authorize('join', $event);
+        $user = $request->user();
+
+        // FIXME: when attempting to join multiple times in a row, only the first attempt triggers the banner.
+        if ($user->cannot('join', $event)) {
+            return redirect()->back()->dangerBanner('Não pode inscrever-se neste evento.');
+        }
 
         $edition = $request->input('edition');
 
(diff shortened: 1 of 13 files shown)


resources/js/Pages/Event.vue
```diff
@@ -72,43 +75,47 @@ const colorPicker = () => {
...
         </section>
         <!-- sign up -->
         <div
             class="flex w-full flex-col items-center gap-4 place-self-center py-24"
         >
             <p class="w-fit text-3xl font-bold text-2023-red">Vamos a isto?</p>
-            <a
-                class="flex cursor-pointer flex-col border border-black bg-2023-teal-dark p-3 px-5 text-center text-xl font-bold text-white shadow-2023-red transition-all hover:shadow-md"
+
+            <PrimaryButton
```
toni-santos, line 124: This is missing a flex and a column orientation

app/Http/Controllers/EventController.php
```diff
@@ -13,4 +15,30 @@ public function show(Event $event)
...
+        $currentEnrollment = $user->usertype->enrollments()->where('edition_id', $edition->id)->first(); // we can safely get only the first one because there should only be one.
+
+        if ($currentEnrollment === null) {
+            return response('No enrollment found', 500);
+        }
+
+        $currentEnrollment->events()->attach($event);
+
+        return redirect()->route('profile.show')->banner("Inscrição no evento '$event->name' realizada com sucesso!");
+    }
```
toni-santos: When I attempted to sign up for an event with an account that has no enrollment I got a "No enrollment screen", we could maybe use laravel's banner to display this alert and redirect the user to the home page.

toni-santos (changes requested):
This is great! Few changes needed.

app/Http/Controllers/EventController.php
Naapperas: The other option is to disable the user from even issuing this request if they are not enrolled, what do you think? Your way is definitely simpler to implement

toni-santos: While not allowing the user at all to sign up for the event may seem more intuitive it can also be confusing. Imo, the banner method acts as a reminder as to how things work. A message reminding them that they must sign up for the edition should suffice and point them in the right direction. Furthermore, if people find that they aren't able to find where to sign up we can also add a "scroll to bottom" type button on the homepage, only if that manifests itself as a problem.

Naapperas:
This should only be merged after #116

app/Http/Controllers/EventController.php
ttoino: I think the best UX wouldn't show an error, that makes the user think they did something wrong. Imo, we shouldn't ever show an action that the user can't take. In this case we should show a button to enroll in the edition.

toni-santos: I didn't consider showing a button for sign up on the edition, that's a good idea.

toni-santos (dismissed):

toni-santos (approved):

PR #145, opened 2023-10-02
Add backend for quests

### LINKED ISSUE #141: Add quest backend
Add a way for a quest to be given to a user through a qr code in their profile.

ttoino:
Closes #141.


Naapperas (dismissed):

app/Http/Controllers/EventController.php
```diff
@@ -39,7 +42,7 @@ public function join(Request $request, Event $event)
 
         // FIXME: when attempting to join multiple times in a row, only the first attempt triggers the banner.
         if ($user->cannot('join', $event)) {
-            return redirect()->back()->dangerBanner('Não pode inscrever-se neste evento.');
+            return redirect()->back()->dangerBanner('Não podes inscrever-se neste evento');
```
Naapperas:
```suggestion
            return redirect()->back()->dangerBanner('Não podes inscrever-te neste evento');
```

app/Http/Controllers/QuestController.php
```diff
@@ -0,0 +1,28 @@
...
+
+        $edition = $request->input('edition');
+        $participant = Participant::firstWhere('quest_code', $request->get('quest_code'));
+        $enrollment = $participant->enrollments()->where('edition_id', $edition?->id)->first();
+
+        Gate::authorize('give', [$quest, $enrollment]);
+
+        $enrollment->quests()->attach($quest);
+
+        return redirect()->back()->banner('Quest atribuída com sucesso!');
```
Naapperas, line 26: Beware of text translations.

app/Providers/AuthServiceProvider.php
```diff
@@ -45,5 +49,24 @@ public function boot(): void
                 $event->enrollments()->count() < $event->capacity // event must not be full
             )
         ));
+
+        Gate::define('give', fn (User $user, Quest $quest, Enrollment $enrollment) => (
+            $enrollment->quests()->where('quest_id', $quest->id)->doesntExist() && // participant must not have the quest
+            ( // either
+                (
+                    $user->isAdmin() && // user is admin and either
```
Naapperas, line 57: Shouldn't this allow for staff to perform this action?

Naapperas (changes requested):

app/Providers/AuthServiceProvider.php
ttoino: Yes, but that's a little hard right now, we should do that in a bigger PR so staff have all the necessary permissions.

app/Http/Controllers/QuestController.php
ttoino: We're not translating any banners right now, maybe #75 can handle that

app/Http/Controllers/EventController.php
ttoino: You're right, mb

Naapperas (approved):

PR #149, opened 2023-10-02
Fix team page

toni-santos:
Remove the border on the team page, and other tweaks.

PR #153, opened 2023-10-02
Normalized slot sizes for images with different aspect ratios

### LINKED ISSUE #152: Normalize profile picture sizes in Team page
Currently, the size of the member slots in the Team page are being affected by the resolution of their profile pictures, leading to this:

image

We should resize/cover the images as necessary in order to ensure the slots are of the same size.

pedrosilva17:
Closes #152 
Self explanatory, simple fix.

Before:
2023-10-02-211933_hyprshot

After:
2023-10-02-212247_hyprshot


pedronunes19 (approved):
This is also done in #149

PR #186, opened 2023-10-07
Fixed events not loading users on Admin edit page

### LINKED ISSUE #183: Editing an Event removes associated speakers and companies

Naapperas:
Closes #183


app/Http/Controllers/CRUDController.php
```diff
@@ -68,6 +69,16 @@ protected function with(): array
...
 
+    /**
+     * The associated relations to load when loading the model.
+     *
+     * @return array<mixed, mixed>
+     */
+    protected function load(): array
+    {
+        return [];
+    }
```
ttoino: What do you think about making this a property, since it doesn't take any arguments?

ttoino (dismissed):

app/Http/Controllers/CRUDController.php
Naapperas: I was following the same principle used on `with`, I can change it but I believe they should follow the same "pattern". Thoughts?

Naapperas:
Also forgot to mention, but as a bonus I moved the relation loading from the EventDay model to the Program page which greatly reduced the number of models loaded when visiting the Home page.

app/Http/Controllers/CRUDController.php
ttoino: The `with` function needs dynamic data, while this one doesn't, but I get what you mean, I'll leave it for you to decide.

Naapperas: Ok, that makes sense. It's not that big of a deal.

ttoino (approved):

ttoino (approved):

PR #189, opened 2023-10-08
Made it so user profile dropdown appears on top of mobiile content dropdown

### LINKED ISSUE #155: Year selection dropdown open when mobile hamburger also open.
Strange behavior on mobile when the user opens the hamburger menu and then opens the year selection dropdown.

Naapperas:
Closes #155

### DIFF
--- a/resources/js/Components/HamburgerMenu.vue
+++ b/resources/js/Components/HamburgerMenu.vue
@@ -67,7 +67,7 @@ onUnmounted(() => {
     </div>
     <div
         v-show="open"
-        class="absolute left-0 top-[5.6rem] z-50 m-0 flex h-screen w-full flex-col bg-2023-teal-dark py-6 text-2xl font-semibold text-2023-bg md:hidden"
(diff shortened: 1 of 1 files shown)


ttoino (approved):

resources/js/Components/HamburgerMenu.vue
```diff
@@ -67,7 +67,7 @@ onUnmounted(() => {
     </div>
     <div
         v-show="open"
-        class="absolute left-0 top-[5.6rem] z-50 m-0 flex h-screen w-full flex-col bg-2023-teal-dark py-6 text-2xl font-semibold text-2023-bg md:hidden"
```
toni-santos, line 70: The `z-50` was added so that the it stays on top of the map on the home page. Removing it might break this behaviour, but I'm not sure.

toni-santos (commented):
Other than the mentioned 👍

resources/js/Components/HamburgerMenu.vue
Naapperas: Oh, I didn't actually test that, nice catch. I'll look into it.

Naapperas: I tested this today and it seemed fine. I might have missed some edge case but for now I'll mark this as resolved.

PR #216, opened 2023-10-19
Home page informative popup

### LINKED ISSUE #198: Add pop-up on edition enrollment explaining gamification
After a user first enrolls, the first time they go somewhere (for example, the Home page), a pop-up should appear explaining them the rules of gamification.

toni-santos:
Add an informative popup to the homepage, informing users of the events informations.
Closes #198

### DIFF
--- a/resources/js/Pages/Home.vue
+++ b/resources/js/Pages/Home.vue
@@ -52,7 +52,7 @@ const formattedDate = (
 <template>
     <AppLayout title="Home">
         <ModalsContainer />
-        <InfoPopup />
+        <InfoPopup v-if="$page.props.auth.user" />
         <a
             v-if="canEnroll"
             href="#enroll-wrapper"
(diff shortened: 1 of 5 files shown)


resources/js/Pages/Home.vue
```diff
@@ -50,6 +51,8 @@ const formattedDate = (
 
 <template>
     <AppLayout title="Home">
+        <ModalsContainer />
```
Naapperas, line 54: Don't we already render this component somewhere else?

toni-santos: This was previously defined in this same file further down, I just moved it up

Naapperas: My bad, I did not see the line deletion below.

Naapperas (dismissed):

Naapperas (approved):

PR #219, opened 2023-10-19
Added support for markdown descriptions

Naapperas:


resources/js/Components/Event/SpeakerInfo.vue
```diff
@@ -94,11 +94,10 @@ const iconColor: Record<string, string> = {
                 </div>
             </div>
             <p
```
ttoino: We should probably not use a `p` here, the markdown will introduce `p`s as children

resources/js/Components/Home/Sponsor.vue
```diff
@@ -51,9 +51,10 @@ defineProps<Props>();
             <span v-else class="text-xl font-bold underline">
                 {{ company?.name }}
             </span>
-            <p class="text-justify max-md:h-96 max-md:overflow-y-scroll">
-                {{ company?.usertype?.description }}
-            </p>
+            <p
```
ttoino: Same as above

ttoino (changes requested):

ttoino (approved):

PR #229, opened 2023-10-21
Added support for sending mails to every participant

Naapperas:
Due to a recent incident, the need to be able to send emails to every participant on demand appeared. This PR implements that.


.env
ttoino, line 1: These changes should be undone

resources/views/emails/participant-notification.blade.php
```diff
@@ -0,0 +1,12 @@
+<x-mail::message>
+    # Introduction
+
+    The body of your message.
+
+    <x-mail::button :url="''">
+        Button Text
+    </x-mail::button>
```
ttoino: This should be replaced with the correct message, right?

app/Mail/ParticipantNotification.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * Get the message envelope.
+     */
+    public function envelope(): Envelope
+    {
+        return new Envelope(
+            subject: 'Participant Notification',
+            bcc: ['nunoafonso2002@gmail.com', 'sirze@live.com.pt']
```
ttoino: This is just temporary, right?

app/Console/Commands/SendMailToParticipants.php
ttoino, line 1: It would be great if we could test the email before sending it, maybe if there is a boolean argument called `test` or something similar, it sends to the admins first, then to the users after its confirmed to be working

resources/views/emails/participant-notification.blade.php
Naapperas: Correct, this is just temporary.

app/Mail/ParticipantNotification.php
Naapperas: Correct.

app/Console/Commands/SendMailToParticipants.php
Naapperas: There are mailing tests in Laravel, is that something we could explore? Or perhaps I misunderstood your comment.

toni-santos (approved):

PR #259, opened 2023-10-22
added support for general quests

### LINKED ISSUE #252: Allow quests that have no stand or event associated
Self-explanatory.

toni-santos:
Closes #252 

Introduces "general quests", that have no associated stall or event.

### DIFF
--- a/app/Http/Controllers/QuestController.php
+++ b/app/Http/Controllers/QuestController.php
@@ -19,6 +19,16 @@ class QuestController extends Controller
         $participant = Participant::firstWhere('quest_code', $request->get('quest_code'));
         $enrollment = $participant->enrollments()->where('edition_id', $edition?->id)->first();
 
+        if ($enrollment === null) {
+            return redirect()->back()->dangerBanner('Participante não inscrito nesta edição!');
+        }
+
+        if (! $quest->requirement) {
+            $enrollment->quests()->attach($quest);
+
+            return redirect()->back()->banner('Tarefa atribuída com sucesso!');
+        }
+
         if (Gate::denies('give', [$quest, $enrollment])) {
             return redirect()->back()->dangerBanner('Não foi possível atribuir a tarefa ao participante!');
         }
--- a/app/Http/Controllers/QuestController.php
+++ b/app/Http/Controllers/QuestController.php
@@ -23,12 +23,6 @@ class QuestController extends Controller
             return redirect()->back()->dangerBanner('Participante não inscrito nesta edição!');
         }
 
-        if (! $quest->requirement) {
-            $enrollment->quests()->attach($quest);
-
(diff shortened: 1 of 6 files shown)


toni-santos:
This PR introduces a migration that must be thoroughly checked.

database/migrations/2023_10_22_134806_nullable_quest_requirement.php
```diff
@@ -0,0 +1,38 @@
...
+    public function up(): void
+    {
+        // This must be separated as drops and creates of the same column cannot be done in the function
+        Schema::table('quests', function (Blueprint $table) {
+            $table->dropMorphs('requirement');
+        });
+
+        Schema::table('quests', function (Blueprint $table) {
+            $table->nullableMorphs('requirement');
+        });
```
ttoino, line 21: This will remove all requirements from existing quests, doing it manually could be a better approach, but we don't have any quests in production so it should be fine (cc @Naapperas)

app/Http/Controllers/QuestController.php
```diff
@@ -19,6 +19,16 @@ public function give(Request $request, Quest $quest)
...
 
+        if ($enrollment === null) {
+            return redirect()->back()->dangerBanner('Participante não inscrito nesta edição!');
+        }
+
+        if (! $quest->requirement) {
+            $enrollment->quests()->attach($quest);
+
+            return redirect()->back()->banner('Tarefa atribuída com sucesso!');
+        }
```
ttoino: These checks should be in the `give` gate.

ttoino (changes requested):

app/Http/Controllers/QuestController.php
toni-santos: We must still check if `enrollment` is null here as the gate must take in a valid `enrollment`, not `null`.

ttoino (approved):

database/migrations/2023_10_22_134806_nullable_quest_requirement.php
Naapperas: I find it weird that these need to be two separate `Schema::table` calls.

Naapperas: Regarding your comment @ttoino, I believe there will be no problem with this. Even if there are quests, it's just a matter of removing them before we migrate the new changes and inserting them afterwards.

Naapperas (approved):

database/migrations/2023_10_22_134806_nullable_quest_requirement.php
toni-santos: > I find it weird that these need to be two separate `Schema::table` calls.

I agree, but, with a single one, I wasn't able to delete and create the morphs.

PR #260, opened 2023-10-22
Change event description type to text

ttoino:


Naapperas (approved):

Naapperas:
Do the HTML forms mark the new nullable fields as required?

ttoino:
> Do the HTML forms mark the new nullable fields as required?

What do you mean?

Naapperas:
> > Do the HTML forms mark the new nullable fields as required?
> 
> What do you mean?

~~Does the FE have the "description/regulation" inputs marked as required or not?~~ Never mind, I only noticed the "step=60" change.

PR #261, opened 2023-10-22
Feature/staff permissions

### LINKED ISSUE #249: Create staff gate
We need a way for users that are assigned to that edition staff to be able to view certain restricted pages (users signed up in an event, ...) and have access to certain features (scan QR codes).

Naapperas:
Closes #249


Naapperas:
There is no way for staff to see which users enrolled into which events because that page is yet to be done, as commented on the original issue.

app/Providers/AuthServiceProvider.php
```diff
@@ -54,13 +56,18 @@ public function boot(): void
...
                 (
-                    $user->isAdmin() && // user is admin and either
                     (
-                        $quest->requirement_type === Event::class &&
-                        $enrollment->events()->where('event_id', $quest->requirement_id)->exists()
-                    ) || // requirement is an event the user has joined
-                    $quest->requirement_type === Stand::class || // or requirement is a stand
-                    $quest->requirement_type === null // or there is no requirement
+                        $user->isAdmin() || // user is admin
+                        $user->isStaff($enrollment->edition) // or user is staff
```
toni-santos, line 61: I think this is a bit redundant, but it also doesn't hurt to be sure.

toni-santos (approved):

app/Providers/AuthServiceProvider.php
Naapperas: Why would this be redundant?

toni-santos: These checks are also made in the pages that use the gate.

Naapperas: Oh yeah, you're right. Better safe than sorry tho.

PR #262, opened 2023-10-23
Implemented product buy action

### LINKED ISSUE #254: Product reservation
There is currently no back-end to reserve/"buy" products.

Naapperas:
Closes #254

### DIFF
--- a/app/Http/Controllers/ShopController.php
+++ b/app/Http/Controllers/ShopController.php
@@ -2,7 +2,6 @@
 
 namespace App\Http\Controllers;
 
-use App\Models\Enrollment;
 use App\Models\Product;
 use App\Models\User;
 use Illuminate\Http\Request;
@@ -17,16 +16,23 @@ class ShopController extends Controller
 
         if ($user && $user->isParticipant()) {
             $isParticipant = $user->isParticipant();
-            $isEnrolled = $user->usertype->enrollments()->where('edition_id', $edition->id)->exists();
-            $points = $user->usertype->enrollments()->where('edition_id', $edition->id)->first()->points;
+            $enrollment = $user->usertype->enrollments()->where('edition_id', $edition->id)->first();
+
+            // This makes one less query to the DB ☝️🤓
+            $isEnrolled = $enrollment !== null;
+            $points = $enrollment?->points;
         }
 
         if ($edition === null) {
             return response('No edition found', 500);
         }
 
+        $products = $edition->products()->orderBy('price')->get()->each(function (Product $product) use ($user) {
+            $product->canBeBought = $user->can('buy', $product);
+        });
+
         return Inertia::render('Shop', [
-            'products' => $edition->products,
+            'products' => $products,
             'points' => $points ?? null,
             'isEnrolled' => $isEnrolled ?? null,
             'isParticipant' => $isParticipant ?? null,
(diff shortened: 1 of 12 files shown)


toni-santos (commented):
Looks good and seems to work, however we should make users aware that products that have been bought cannot be bought again. This shouldn't be too hard and could be done in this PR.

app/Http/Controllers/ShopController.php
```diff
@@ -29,4 +32,42 @@ public function show(Request $request)
...
+
+        if ($currentEnrollment === null) {
+            return redirect()->back()->dangerBanner('You are not enrolled in this edition');
+        }
+
+        $points = $currentEnrollment->points;
+
+        if ($points < $product->price) {
+            return redirect()->back()->dangerBanner('You do not have enough points');
+        }
```
ttoino: This should be done in a gate or policy, and the error message(s) should be in portuguese

app/Http/Controllers/ShopController.php
```diff
@@ -29,4 +32,42 @@ public function show(Request $request)
...
+
+        $currentEnrollment->decrement('points', $product->price);
+        $currentEnrollment->products()->attach($product->id);
+
+        return redirect()->back()->banner('Product bought');
```
ttoino: This should be in portuguese

app/Http/Controllers/ShopController.php
```diff
@@ -29,4 +32,42 @@ public function show(Request $request)
...
+            return redirect()->back()->dangerBanner('You are not enrolled in this edition');
+        }
+
+        $points = $currentEnrollment->points;
+
+        if ($points < $product->price) {
+            return redirect()->back()->dangerBanner('You do not have enough points');
+        }
+
+        $currentEnrollment->decrement('points', $product->price);
```
ttoino: Don't we have a trigger for this?

resources/js/Components/Shop/ShopItem.vue
```diff
@@ -58,8 +65,7 @@ defineProps<Props>();
                 Confirmar compra de <b>{{ product.name }}</b> por
                 {{ product.price }}?
             </p>
-            <!-- TODO: BACKEND CONNECTION -->
-            <PrimaryButton>Comprar</PrimaryButton>
+            <PrimaryButton @click="buyProduct()">Comprar</PrimaryButton>
```
ttoino, line 70: We should check the aforementioned gate here, so the user doesn't see the button if they can't buy the product.

ttoino (changes requested):

app/Http/Controllers/ShopController.php
toni-santos: Also, isn't in portuguese

Naapperas: We do not @ttoino, should I create it?

Naapperas: Will create it.

resources/js/Components/Shop/ShopItem.vue
Naapperas: :point_up: :nerd_face: "aforementioned"

app/Http/Controllers/ShopController.php
```diff
@@ -15,18 +17,48 @@ public function show(Request $request)
         if ($user && $user->isParticipant()) {
             $isParticipant = $user->isParticipant();
             $isEnrolled = $user->usertype->enrollments()->where('edition_id', $edition->id)->exists();
-            $points = $user->usertype->enrollments()->where('edition_id', $edition->id)->first()->points;
+
+            $points = $isEnrolled ? $user->usertype->enrollments()->where('edition_id', $edition->id)->first()->points : null;
```
ttoino: This queries the database for the enrollment twice

ttoino (dismissed):

toni-santos (approved):

PR #315, opened 2023-10-27
Started adding logs to application

### LINKED ISSUE #311: Add logging.
We should log every action taken on the platform.

Naapperas:
Closes #311


app/Actions/Fortify/CreateNewUser.php
```diff
@@ -27,13 +29,17 @@ public function create(array $input): User
...
+        $data = [
             'name' => $input['name'],
             'email' => $input['email'],
             'password' => Hash::make($input['password']),
             'usertype_id' => '0',
             'usertype_type' => Participant::class,
-        ]);
+        ];
+
+        Log::info('Creating user with input: {input}', ['input' => Json::encode($data, true)]);
```
ttoino: I don't think logging a user's password is a great idea 🥴

app/Http/Controllers/EventController.php
```diff
@@ -52,26 +53,45 @@ public function show(Request $request, Event $event)
...
 
         if ($edition === null) {
             return response('No edition found', 500);
         }
 
+        /** @var User|null */
+        $user = $request->user();
+
+        if ($user === null) {
+            return redirect()->back()->dangerBanner('Tens que ter sessão iniciada para fazer esta ação');
```
ttoino: I don't like this error message, it sounds too robotic

app/Http/Controllers/QuestController.php
```diff
@@ -23,11 +24,29 @@ public function give(Request $request, Quest $quest)
...
         }
 
-        if (Gate::denies('give', [$quest, $enrollment])) {
+        /** @var User|null */
+        $user = $request->user();
+
+        if ($user === null) {
+            Log::warning('Unauthenticated user attempted to give quest to participant');
+
+            return redirect()->back()->dangerBanner('Tens que estar autenticado para efetuar esta ação');
```
ttoino: Same as above

ttoino (changes requested):

ttoino:
Actually, now that I think about it, are those error messages when the user is not logged in even required? The routs have the fortify middleware anyway

app/Actions/Fortify/CreateNewUser.php
Naapperas: :sob: my bad

app/Http/Controllers/EventController.php
Naapperas: :robot:

Naapperas: Do you have any suggestions? @toni-santos suggested `Inicia sessão para fazer esta ação` :musical_note:

app/Http/Controllers/QuestController.php
Naapperas: :robot:

Naapperas:
> Actually, now that I think about it, are those error messages when the user is not logged in even required? The routs have the fortify middleware anyway

routs

Naapperas:
> Actually, now that I think about it, are those error messages when the user is not logged in even required? The routs have the fortify middleware anyway

Perhaps these specific routes will never fall in that branch but better be safe than sorry.

CiscoPr (approved):

ttoino (approved):

PR #319, opened 2023-10-29
Feature/competition leaderboards

Naapperas:


pedrosilva17:
Any screenshots? 🙏

Naapperas:
> Any screenshots? 🙏

image

image

image

I just found an issue with the podium not updating when we change the current competition, gotta check into that.

pedrosilva17:
The frontend looks fine, no glaring issues with the code either (from what I've seen). Fix that issue and it should be good to go

app/Http/Controllers/CompetitionController.php
```diff
@@ -39,8 +39,20 @@ public function show(Request $request, Competition $competition)
             }
         }
 
+        $leaderboard = $competition->teams()->orderBy('points', 'desc')->take(3)->get();
```
ttoino: Can't we do this on the client? It would save a db query

app/Http/Controllers/CompetitionTeamCRUDController.php
ttoino, line 1: We should probably also add a way to associate team members

resources/js/Components/Podium.vue
```diff
@@ -358,5 +394,35 @@
...
+            </div>
+            <div
+                class="align-center text-wrap flex flex-col items-center truncate px-1 text-lg"
+            >
+                <span class="whitespace-normal font-bold">{{
+                    leaderboard[2].name
+                }}</span>
+                <span class="">{{ leaderboard[2].points }}</span>
+            </div>
+        </div>
```
ttoino: I think it's ok to not have this on mobile, as it looks a little weird, and all teams are present in the list below

resources/js/Pages/Competition.vue
```diff
@@ -75,7 +85,36 @@ const formattedDate = (
...
+                    </span>
+                    <div
+                        class="flex w-3/4 flex-col border border-black shadow-lg shadow-2023-teal md:w-1/2"
+                    >
+                        <div
+                            v-for="team in competition.teams ?? []"
+                            :key="team.id"
+                            class="inline-flex w-full justify-between p-4 text-lg even:bg-2023-orange even:bg-opacity-20"
+                        >
+                            {{ team.name }} - {{ team.points }}
```
ttoino: I think we should also show the team image and the points should be below the name or justified to the right

ttoino (changes requested):

ttoino (approved):

PR #320, opened 2023-10-30
Starter working on scanning QR codes from event page

### LINKED ISSUE #282: QR code scanning in event page
Currently scanning QR codes can only be done through the profile, this makes it so checking if users are enrolled in an event and scanning their codes is split into 2 separate places, making it harder to resolve issues in critical situations.

Naapperas:
Closes #282


pedronunes19 (approved):
bloat

PR #323, opened 2023-10-30
Fixed shop item disabling

Naapperas:


app/Http/Controllers/ShopController.php
```diff
@@ -31,7 +31,10 @@ public function show(Request $request)
         }
 
         $products = $edition->products()->orderBy('price')->get()->each(function (Product $product) use ($user) {
-            $product->canBeBought = $user ? $user->can('buy', $product) : true;
+            $product->canBeBought =
+                $product->stock <= 0
```
CiscoPr, line 35: why don't you use the '==' operator instead of the '<=' here? wouldn't it make more sense?

Naapperas: `<=` already accounts for `==`: I added this just as a safety measure.

pedronunes19 (approved):

CiscoPr (approved):

